PR #1, opened 2018-05-01
README.md - Point to installation gist

totten:
@jackrabbithanna @dsnopek, my recollection is that we've generally started sharing the gist as the installation instructions because this gets the problem of hidden class-loader conflicts.

In the (hopefully-not-too-distant) future, as we remove barriers and get the installation procedure more polished, we'll want to add a chapter to the sysadmin guide. But for now, it feels like we should be referencing the instructions that folks are using.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,14 +5,7 @@ This is alpha version of the integration module required to allow CiviCRM and Dr
 
 It it currently verified to work against Drupal 8.0.5 and CiviCRM 4.7.3.
 
-Installation Steps
-------------------
+Installation
+------------
 
-- Download the latest Drupal 8 release: (from https://www.drupal.org/project/drupal).
-- Install Drupal 8 (see https://api.drupal.org/api/drupal/core!INSTALL.txt/8 for more information).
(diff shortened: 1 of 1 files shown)


dsnopek:
Hm. The Gist is really only part of the instructions - the part meant to be replaced at some point by `composer require civicrm/civicrm-core`. There's some post steps in the blog post:

https://www.mydropwizard.com/blog/how-install-civicrm-drupal-8-and-why-choose-it-over-pure-drupal-crm (See "The full, manual process")

Maybe these instructions should just be copied into the README.md file to make them "official"?

JoeMurray:
+1

Joe Murray from mobile

borisson:
I disagree with changing the text to another place (either the gist or the blogpost by @dsnopek). Having all the steps **here** makes it easier for others to update the information when needed.

eileenmcnaughton:
@totten this is stale should it be rebased or closed?

mlutfy:
The README has already been updated, so feels OK to close.

PR #3, opened 2018-05-17
dev/drupal#22: implement support for Drupal8 views and multilingual C…

mlutfy:
This is a (mostly copy-pasted) port of the Drupal7 views code.

### DIFF
--- a/civicrm.views_execution.inc
+++ b/civicrm.views_execution.inc
@@ -1,0 +1,32 @@
+<?php
+
+use Drupal\views\ViewExecutable;
+use Drupal\views\Plugin\views\query\QueryPluginBase;
+
+/**
+ * Implements hook_views_query_alter().
+ */
+function civicrm_views_query_alter(ViewExecutable $view, QueryPluginBase $query) {
+  \Drupal::service('civicrm')->initialize();
+
+  // check if we are in multilingual mode, otherwise return
+  // TODO: should be a simple call - is_multiligual ?
+  $domain = new CRM_Core_DAO_Domain();
+  $domain->find(TRUE);
+
+  $multilingual = (bool) $domain->locales;
+
+  if ($multilingual) {
+    global $dbLocale;
+    $columns = CRM_Core_I18n_SchemaStructure::columns();
+
+    // TODO: for better performance, loop on $query->fields instead
+    foreach ($columns as $table => $hash) {
+      foreach ($hash as $column => $type) {
+        if (array_key_exists("{$table}_{$column}", $query->fields)) {
+          $query->fields["{$table}_{$column}"]['field'] = "{$column}{$dbLocale}";
+        }
+      }
+    }
+  }
+}


civicrm.views_execution.inc
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Implements hook_views_query_alter().
+ */
+function civicrm_views_query_alter(ViewExecutable $view, QueryPluginBase $query) {
+  \Drupal::service('civicrm')->initialize();
+
+  // check if we are in multilingual mode, otherwise return
+  // TODO: should be a simple call - is_multiligual ?
+  $domain = new CRM_Core_DAO_Domain();
```
seamuslee001, line 14: @mlutfy does this need a \ at the start to say we should be in the root namespace?

mlutfy: I don't think it's required/suggested? It seems to work in both cases, but grepping around, I don't see `new \CRM`. The only place I see it, is under `civicrm-core/Civi/`.

Correct me if I'm wrong, but googling around, people seem to suggest avoiding it if not required, for code readability. It should only be used when there is a risk of a namespace clash between the local class and global scope.

(sorry for the slow response, had not seen the email notification on this thread)

jackrabbithanna:
Just wanted to let you'll know that the latest CiviCRM Entity 8.x-3.x-dev version includes Views support....Not telling you what to do, but the "Core Views Integration" could be handled by that module....

mlutfy:
I opened https://lab.civicrm.org/dev/drupal/issues/28 to discuss Views deprecation from Core. Seems like a good idea.

Nonetheless, until that happens, can we merge this? I'll also start experimenting with CiviCRM Entity and test how it handle multi-lingual.

jackrabbithanna:
I'd say merge it if it works...It is likely CiviCRM Entity will need some portion of Views handler code ported over from the core Views integration module to reach feature parity....

It will likely be a lot of copy and pasting, not everything needs to be rewritten, but HOW it is integrated may be changed a little bit...

We'll see what the feedback is on #28 before a large amount of time is spent on this, but now is the time to discuss a switch over as a community, before any official releases go out...

JoeMurray:
mlufty could you try creating views with civi d8 entity and entity views integration? It worked for us for simple cases and it would be good to consolidate development efforts on one approach. Leveraging more D8 code and supporting less civi specific code seems wiser given our linited resources.

civicrm.views_execution.inc
borisson: It's not required, but it is more explicit and that's a great thing. It's also marginally faster.

borisson:
It seems like the conclusion there is to remove it from this repo. So I think this PR should be closed and a new one opened to remove all the views integration from this module?

mlutfy:
It's not on the top of my list, but I'll close the PR for now. We can continue discussion here: https://lab.civicrm.org/dev/drupal/issues/22

mlutfy:
For anyone stumbling on this issue: CiviCRM Entity does handle multilingual views without any issues.

PR #4, opened 2018-06-28
Added missing civicrm initialize

kewljuice:
Without civicrm-initialize Drupal Views throws error: Class 'CRM_Core_BAO_LocationType' not found!
missing_civicrm_initialize


colemanw:
Looks like a useful fix. Thanks @kewljuice!

PR #5, opened 2018-08-07
Replaced _getStandardTypeFromCustomDataType for civicrm-core 5.4.0

kewljuice:
I replaced _getStandardTypeFromCustomDataType with CRM_Utils_Array::value($dao->data_type, CRM_Core_BAO_CustomField::dataToType()) because the function is removed in civicrm-core 5.4.0.


colemanw:
Thanks @kewljuice !

PR #6, opened 2018-08-20
Commented out Drupal7 function

kewljuice:
Commented out Drupal7 function 'variable_del'


colemanw:
Is any equivalent step needed for D8 or is it completely superfluous? In that case we can just remove the line and related comments.

borisson:
The concept of variables doesn't exist in drupal 8 anymore.

This line is the only mention of `civicrm_class_loader` in the codebase, so we should probably delete the line and the comment next to it.

borisson:
Looks great now!

PR #8, opened 2018-08-24
Add dedicated exception

borisson:
Uses a specific exception class instead of the wrong ConfigException class, which should only be used with configuration synch issues.


kewljuice:

PR #9, opened 2018-08-24
Use DI to init CiviCRM in the permissions builder

borisson:


kewljuice:

PR #10, opened 2018-08-24
Remove unneeded file

borisson:
This file was never being executed because it of the `.php.disabled` extension.


kewljuice:
I've tested this patch, looks like everything works fine without this file!

PR #11, opened 2018-08-24
Only synchronize users when they are logged in

borisson:


kewljuice:

PR #12, opened 2018-08-30
Improve active theme

borisson:
Improves the readability of the active theme switcher, we should also return quicker in some cases, making for a marginally faster result.

The commits should probably be squashed when merging this PR.


mlutfy:
This seems OK to me. I did a bit of testing on a production site and did not see any errors. Thank you @borisson !

PR #14, opened 2018-08-30
Refactor path processor and document behavior with unit-test.

borisson:
Refactor path processor and document behavior with unit-test. This should also lead to a marginal performance improvement.

### DIFF
--- a/civicrm.services.yml
+++ b/civicrm.services.yml
@@ -10,5 +10,6 @@ services:
       - { name: breadcrumb_builder, priority: 1002 }
   civicrm.path_processor:
     class: Drupal\civicrm\PathProcessor\CivicrmPathProcessor
+    arguments: ['@civicrm']
     tags:
       - { name: path_processor_inbound, priority: 250 }
--- a/src/PathProcessor/CivicrmPathProcessor.php
+++ b/src/PathProcessor/CivicrmPathProcessor.php
@@ -12,45 +12,65 @@ use Drupal\civicrm\Civicrm;
 class CivicrmPathProcessor implements InboundPathProcessorInterface {
 
   /**
+   * The CiviCRM service.
+   *
+   * @var \Drupal\civicrm\Civicrm
+   */
+  protected $civiCRM;
+
+  /**
+   * CivicrmPathProcessor constructor.
+   *
+   * @param \Drupal\civicrm\Civicrm $civicrm
+   *   The CiviCRM service.
+   */
+  public function __construct(Civicrm $civicrm) {
+    $this->civiCRM = $civicrm;
+  }
+
+  /**
    * {@inheritdoc}
    */
   public function processInbound($path, Request $request) {
-    // If the path is a civicrm path.
-    if (strpos($path, '/civicrm/') === 0) {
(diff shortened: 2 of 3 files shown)


colemanw:
That's great that you've added a test @borisson. I'm not sure though if our CI is set up to work with Drupal 8's testing framework. @totten @mlutfy what do you think?

borisson:
You can run the tests with `phpunit` locally to verify that this is working. When we are closing to being able to install this module & CiviCRM trough composer we can easily add travis to this repository. 

If you say that these are not being tests, we should probably also remove the simpletests in this repo as well?

colemanw:
@borisson are you on https://chat.civicrm.org/civicrm/channels/drupal ? I just mentioned you there in relation to the great work you've been doing in this repo.

totten:
@demeritcowboy @seamuslee001 @colemanw  - I stumbled across a bug (https://lab.civicrm.org/dev/core/-/issues/4478) which is fixed by this PR. I wonder if there's any particular reason why it's been left open?

---

Aside: The PR introduces `Drupal\Tests\UnitTestCase`, which we haven't been using. So I get this:

```
[bknix-max:~/bknix/build/d10/web/modules/contrib/civicrm] phpunit8 tests/src/Unit/CivicrmPathProcessorTest.php 

Fatal error: Uncaught Error: Class "Drupal\Tests\UnitTestCase" not found in /home/totten/bknix/build/d10/web/modules/contrib/civicrm/tests/src/Unit/CivicrmPathProcessorTest.php on line 15
```

 (*Comparison: Integration tests on [Civi-D7](https://github.com/civicrm/civicrm-drupal/tree/7.x-master/tests/phpunit), [Civi-BD](https://github.com/civicrm/civicrm-backdrop/tree/1.x-master/tests/phpunit), and [Civi-WP](https://github.com/civicrm/civicrm-wordpress/tree/master/tests/phpunit) are written in `e2e` style. But maybe there are trade-offs. I don't know `Drupal\Tests\UnitTestCase` well enough to differentiate.*)

demeritcowboy:
I'm guessing you need to use drupal's bootstrap (i.e. with `phpunit -c web/core/tests/bootstrap.php`). I have a modified version [here](https://github.com/demeritcowboy/civix/blob/mink/src/CRM/CivixBundle/Resources/views/Code/phpunit-boot-mink.php.php) that is intended to be run from an extension and so adds civi classpaths which you might not need, but you might be interested in at least the part that handles 9/10 backwards incompatibility: https://github.com/demeritcowboy/civix/blob/c2f81bb25938543650363be372380a6d7981ebf5/src/CRM/CivixBundle/Resources/views/Code/phpunit-boot-mink.php.php#L201-L206 (and remove the `use` at the top).

Also when you install drupal, you might need to add the [drupal/core-dev package](https://github.com/SemperIT/CiviCARROT/blob/08d372df2c340171a9790607eb626ed38c46eff0/.github/workflows/main.yml#L91) ("pinned" or not).

demeritcowboy:
This has file conflicts. It may still have merit but needs revisiting after #97, so going to close.

PR #15, opened 2018-09-03
 Use any version of CiviCRM, greater than 5.4.0

kewljuice:
So we can install CiviCRM & Drupal 8 via composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
   ],
   "require": {
     "composer/installers": "^1.0",
-    "civicrm/civicrm-core" : "4.7.*"
+    "civicrm/civicrm-core" : ">=5.4.0"
   },
   "extra": {
     "installer-name": "civicrm"


kewljuice:
I think it's interesting to create a tag 5.4 because there is CiviCRM 5.4.x specific code in the repository.

eileenmcnaughton:
@colemanw thoughts on this? Seems fine to me to keep incrementing the version alongside civi version as we are a long way from the point where we expect them to run somewhat independently

colemanw:
Our releaser script really ought to be tagging this repo. @totten is that on your radar?

JoeMurray:
@monish pls review

eileenmcnaughton:
I don't see much reason NOT to merge this & "here is CiviCRM 5.4.x specific code in the repository" is compelling - merging

PR #16, opened 2018-10-30
fix logout on user create

wannesderoy:
Current user get's logged out when creating a Drupal account for a CiviCRM contact.
by removing the sync and just initializing CiviCRM this is fixed.

### DIFF
--- a/civicrm.user.inc
+++ b/civicrm.user.inc
@@ -19,7 +19,7 @@ function civicrm_user_login(AccountInterface $account) {
  */
 function civicrm_user_insert(AccountInterface $account) {
   $civicrm = \Drupal::service('civicrm');
-  $civicrm->synchronizeUser($account);
+  $civicrm->initialize();
 
   // As per CRM-7858, the email address in CiviCRM isn't always set
   // with a call to synchronize(). So we force this through.
--- a/civicrm.user.inc
+++ b/civicrm.user.inc
@@ -20,6 +20,11 @@ function civicrm_user_login(AccountInterface $account) {
 function civicrm_user_insert(AccountInterface $account) {
   $civicrm = \Drupal::service('civicrm');
   $civicrm->initialize();
+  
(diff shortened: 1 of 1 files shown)


mlutfy:
This patch does avoid getting logged out, but it seems to break the sync to the CiviCRM contact record?

Ex:

- Create a new Drupal8 user
- Click on the "Name and address" tab of the user, it should display the first/last name of the user (pulled from CiviCRM)
- Go to CiviCRM itself, find the contact that was created: it should be linked to the drupal8 user, and should also have their first/last name and email.

With this patch, CiviCRM created a contact record, but it did not link it to the drupal user.

wannesderoy:
I only noticed the issue when creating a Drupal user starting from the CiviCRM contact, not the other way around. I will take a look and search for a solution to ensure the sync.
Thanks for the feedback @mlutfy

wannesderoy:
I found that the synchronize method in the CRM_Core_BAO_UFMatch class has the following code:
```
    // reset the session if we are a different user
    if ($ufID && $ufID != $userSystemID) {
       $session->reset();
       ...
```
so running the sync is never going to work properly I think. The `civicrm_user_insert` hook should only try to create a new CiviCRM contact and match the Drupal en CiviCRM user/contact.

LarsSchmidtZ:
I had the nearly the same problem as described below:

We have D8.6.2 with civicrm 5.6.1 up an running but we ran into Problems when creating new users in D8 Admin area. The creating Admin is logged out after creating the user because the session key is killed. Deactivating Profiles lead to the known error invalid session key.


The patch by wannesderoy worked for me. The Link to the Drupal user is also in the dataset of Civicrm and users looked to be synced correctly.

mlutfy:
I did more testing and it works great. Thank you @wannesderoy for the patch, and @LarsSchmidtZ for the testing!

wannesderoy:
Thanks all for helping me reach a proper solution and merging into master!

JoeMurray:
Thanks!

PR #17, opened 2018-11-26
Add core resources before rendering content

colemanw:
The other 4 CMSs all behave this way, so this change is for consistency.

See https://lab.civicrm.org/dev/drupal/issues/32

### DIFF
--- a/src/Civicrm.php
+++ b/src/Civicrm.php
@@ -78,6 +78,9 @@ class Civicrm {
    */
   public function invoke($args) {
     $this->initialize();
+    
+    // Add CSS, JS, etc. that is required for this page.
+    \CRM_Core_Resources::singleton()->addCoreResources();
 
     // CiviCRM will echo/print directly to stdout. We need to capture it so that
     // we can return the output as a renderable array.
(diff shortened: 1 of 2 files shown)


monishdeb:
Thanks @colemanw for working in this issue. Here's what I did to test this patch:
1. Applied the patch to civicrm D8 module
2. Installed uk.squiffle.kam 

But the civicrm menubar didn't show up. In the console I get 
```
https://www.d8-demo.com/civicrm/ajax/navmenu?c=tdpc64GO&l=en_US 404 (Not Found)
```
Did I missed any step?

colemanw:
Hey @monishdeb I had the same problem, which I tracked down to the new menu path (`civicrm/ajax/navmenu`) being registered in the *CiviCRM* menu system but not in the *Drupal* menu system. The solution was to clear Drupal caches after enabling the extension.

I've filed an issue for that: https://lab.civicrm.org/dev/drupal/issues/37

monishdeb:
Oops my bad. It worked like a charm:

screen shot 2018-11-27 at 6 35 11 pm

screen shot 2018-11-27 at 6 32 52 pm

Also ensured that it doesn't cause any unintended behavior in Civi w/o uk.squiffle.kam

I am happy with this PR, merging now.

PR #18, opened 2018-11-30
Add tray for improved menubar integration

colemanw:
This is to support the (soon-to-be-in-core) [KAM](https://github.com/aydun/uk.squiffle.kam) extension.

drupal8

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -71,7 +71,7 @@ function civicrm_toolbar() {
   // permissions can control it's visibility.
   $items['civicrm'] = [
     '#cache' => [
-      'contexts' => ['user.permissions'],
+      'contexts' => ['user.permissions', 'url.path'],
     ],
   ];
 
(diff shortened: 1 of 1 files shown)


civicrm.module
```diff
@@ -86,11 +86,23 @@ function civicrm_toolbar() {
...
-            'class' => ['toolbar-item'],
           ],
         ],
       ],
     ];
+    // Add tray if we are on a CiviCRM screen
+    $path = ltrim(\Drupal::service('path.current')->getPath(), '/');
+    if (strpos($path, 'civicrm') === 0) {
+      $items['civicrm']['tray'] = [
+        '#heading' => t('Test tray'),
```
JoeMurray: Should this be removed from production code?

JoeMurray:
What is a tray in this context?

civicrm.module
colemanw: Removed

colemanw:
@JoeMurray it makes the CiviCRM menubar act like a Drupal taskbar tray. See screencast.

JoeMurray:
Okay, watching the screencast about 10 times I eventually noticed that the CiviCRM entry on the top Drupal menu changes into a CiviCRM icon beside a menu hamburger when the CiviCRM menu disappears because the window gets too narrow. Got it. Nice.

JoeMurray:
@monishdeb could we try this for our D8 site with the menuing that is giving us such grief.

dsnopek:
The result and the code to accomplish it look quite similar to Roundearth :-)

colemanw:
Thanks @JoeMurray and @dsnopek. Your testing and feedback of the KAM extension is appreciated.
@monishdeb when you test this out please leave a comment on the KAM PR also.

monishdeb:
Tested, working fine in clean D8 instance. Merging now.

PR #19, opened 2019-02-18
Improved user/contact insert

kewljuice:
- Resolves double creation from CiviCRM contact to Drupal user
- Resolves issue when creating Drupal users from migrate or drush


mlutfy:
Hi! Based on code review, this looks like a nice improvement.

Could you provide a bit more details on how to reproduce the errors?

- How to trigger the duplicate contact creation?
- What kind of errors were there with migrate?

kewljuice:
It seems that creating Drupal users via migrations no longer causes problems, but if you create a new individual contact in CiviCRM and want to create a Drupal user via actions on the contact form CiviCRM gives an error message.

create_user_record

error

demeritcowboy:
I think both issues here have now been solved? There was a situation where the "Create User Record" caused an error but it was addressed in https://github.com/civicrm/civicrm-core/pull/17769

mlutfy:
Closing based on @demeritcowboy 's review. Feel free to reopen if it's not the case.

PR #20, opened 2019-03-06
Added views filter for CiviCRM date type.

kewljuice:


mlutfy:
Hi @kewljuice, sorry for the slow response, but for Drupal8 we should aim to remove the views integration from CiviCRM core and instead rely on entity_civicrm (see: https://lab.civicrm.org/dev/drupal/issues/39).

PR #21, opened 2019-03-20
Fixed breadcrumb caching issues

mfrausto-mdw:
Resolves breadcrumb issues cache related.


sunilpawar:
Tested this PR.

For Url : `civicrm/admin/options/gender?reset=1`
**Before path** : There was no breadcrumb available other than `Home >> CiviCRM` .

**After patch** : Breadcrumb Path : `Home >> CiviCRM >> Administer CiviCRM >> Option Groups `

tested other Urls 

> civicrm/admin/options/acl_role?reset=1
> civicrm/pledge/search?reset=1

Each Url showing correct breadcrumb path

CC : @seamuslee001

seamuslee001:
merging as per review

PR #22, opened 2019-03-22
Update README.md to remove out-dated information and refer to Composer project & plugin

mydropwizard-support:
So, the current README is hopelessly out-of-date: the information is incorrect (you can't install a Drupal 8 site following those instructions) and misleading (it's probably more like 'beta' quality than 'alpha', it works with much newer Drupals and CiviCRMs and actually won't with the versions mentioned there, etc).

There was an attempt to update this in PR #1, However, there wasn't agreement on linking out to my article and Gist.

I don't want to be totally presumptuous, but here's a PR that links out to the Composer project & plugin that I made and wrote about recently:

https://www.mydropwizard.com/blog/better-way-install-civicrm-drupal-8

Compared to the old instructions and Gist, this is quite a lot easier to use. Also, unlike those, they are Open Source repos on GitLab that others can fork and contribute to.

Please let me know what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,34 +1,30 @@
 CiviCRM Drupal 8 Module
 =======================
 
-This is alpha version of the integration module required to allow CiviCRM and 
-Drupal 8 to work together. This is not the stable version yet.
+This is a beta version of the integration module required to allow CiviCRM and 
+Drupal 8 to work together.
 
-It it currently verified to work against Drupal 8.0.5 and CiviCRM 4.7.3.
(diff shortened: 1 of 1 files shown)


dsnopek:
Sorry, I didn't realize I was logged in as the wrong account. :-) This PR is by @dsnopek

colemanw:
Merging as this is an improvement.

dsnopek:
Thanks! :-)

PR #23, opened 2019-07-02
Migrate CivicrmHelper::parseUrl() to CRM_Utils_System_Drupal8::parseUrl()

totten:
This fixes a bug which manifests during CLI installation. It corrects a dependency issue by
migrating an install-critical function from `civicrm-drupal-8` to `civicrm-core`.

This PR is one of two changes.  The first adds `parseUrl()` to `civicrm-core`'s `CRM_Utils_System_Drupal8`; the second removes `parseUrl()` from `civicrm-drupal-8`'s `CivicrmHelper`. (The core PR should be merged first.)

For full discussion of "Before/After/Technical Details", see https://github.com/civicrm/civicrm-core/pull/14696


eileenmcnaughton:
This makes sense given the core pr is now merged

homotechsual:
Merging this PR breaks installs/upgrades of CiviCRM on Drupal 8 which rely on cloning this repo for installs. This is an unversioned repo so current upgrades get these changes but are then missing the other half of the changes in civicrm/civicrm-core#14696

mlutfy:
I opened this issue for the tagging: https://lab.civicrm.org/dev/release/issues/9

totten:
__Short recommendation__: If you're using the roundearth template and not doing development, then I'd suggest updating your root project's `composer.json` to pin both `civicrm-drupal-8` and `civicrm-core` to roughly the same version (e.g. current stable 5.16).

```
composer require civicrm/civicrm-core:~5.16 civicrm/civicrm-drupal-8:5.16.x-dev
```

Going forward, you would bump those two versions up and down in tandem.

(more details to come in another comment)

totten:
A bit more detail:

* [This part of the roundearth project template](https://gitlab.com/roundearth/drupal-civicrm-project/blob/1bff742a/composer.json#L24) means that it's tracking `civicrm-drupal-8@dev-master`. That makes good sense for development - but probably not as much for production.
* PR #23 left `composer.json` in a stale condition. Compare:
    * PR #23 depends on https://github.com/civicrm/civicrm-core/pull/14696, which will go out in v5.18. 
    * However, `composer.json` declares a requirement on `civicrm-core >=5.4.0`. This is now wrong since we need v5.18.
* Intuitively, one wants to have `civicrm-core` and `civicrm-drupal-8` always use corresponding versions. (If one is `5.16.0`, then the other should be `5.16.0`. If one is `5.17.x-dev`, then the other is `5.17.x-dev`. If one is `dev-master`, then the other is `dev-master`.) Unfortunately, there are a few quirks which don't allow this to work automatically (at time of writing):
    * (As @mlutfy pointed out) The repo `civicrm-drupal-8` doesn't have tagged versions. It only has branches. https://lab.civicrm.org/dev/release/issues/9. So you *cannot* track stable releases (even when that's appropriate).
    * The `roundearth/civicrm-composer-plugin` only works with tagged versions of `civicrm-core` - and never works with branches. So you *cannot* track branches (even when that's appropriate).
* The following command is a short-term work-around. It pins `civicrm-core` to a tagged release (to address the plugin quirk), and it pins `civicrm-drupal-8` to a branch (to address the lack of tags), and it ensures that the two packages are roughly the same (5.16).

    ```
    composer require civicrm/civicrm-core:~5.16 civicrm/civicrm-drupal-8:5.16.x-dev
    ```

homotechsual:
Do we have (or are we planning) a CiviCRM "Official" composer plugin/project workflow which is maintained in tandem with the project/releases?

totten:
> Do we have (or are we planning) a CiviCRM "Official" composer plugin/project workflow which is maintained in tandem with the project/releases?

 @MikeyMJCO Yes and no. I think there should ultimately be two small and orthogonal plugins:

* [composer-downloads-plugin](https://github.com/civicrm/composer-downloads-plugin): Aimed at upstream projects (e.g. `civicrm-core`) - it lets them automatically pull down oddball resources. (For `civicrm-core`, that's [bower_components](https://github.com/civicrm/civicrm-core/pull/15044) and [packages](https://lab.civicrm.org/dev/core/issues/1209). For the Mosaico extension, that would be the JS/CSS build.)
* __civicrm-asset-plugin__: Aimed at downstream/root projects - it lets them automatically copy JS/CSS assets to their public/web folder. It has heuristics to pick a decent default output path on common project structures (e.g. D8 tarball and drupal-composer/drupal-project). But they don't strictly need to use it - they could just put everything on the web, tweak-out the `installer-paths`, or use some other file-sync plugin.

Everything else that's currently in the scope of GenCode or of civicrm-composer-plugin -- e.g. generating SQL files, handling bootstrap for `extern` scripts -- shall be patched into irrelevance so that out-of-the-box composer workflows are, eh, working.

(To be clear... that should not detract from the value of providing a project template. The template makes it *much* easier to get started, esp for a newcomer. It just means (a) that the template is thinner and (b) when people go off-script by making their own projects or by changing the templates, the behavior is still well-defined.)

PR #25, opened 2019-07-05
Fixed Error when saving a filter with a pseudoconstant

kainuk:
*Versions*
- Drupal 8.7.3
- CiviCRM 5.14.2

*Problem*
Its possible to create views that filter on pseuconstants, for example gender or contact_type. For example:
pr-optionGroups
A view with this configuration cannot be saved. Drupal generates the following error:

`The website encountered an unexpected error. Please try again later.</br></br><em class="placeholder">InvalidArgumentException</em>: The configuration property display.default.display_options.filters.contact_type.value.Individual doesn&#039;t exist. in <em class="placeholder">Drupal\Core\Config\Schema\ArrayElement-&gt;get()</em> (line <em class="placeholder">76</em> of <em class="placeholder">core/lib/Drupal/Core/Config/Schema/ArrayElement.php</em>). <pre class="backtrace">Drupal\Core\Config\StorableConfigBase-&gt;castValue(&#039;display.default.display_options.filters.contact_type.value.Individual&#039;, &#039;Individual&#039;) (Line: 211) ...`

This PR fixes this problem

*Technical Details*
A comparable error is described on the blog article https://www.computerminds.co.uk/articles/got-config-schema-error-saving-view .

PseudoConstants are not processed by `in_operator` but by a subclass `civicrm_in_operator`. Functionality is coded in CivicrmInOperator. This class ensures the correct startup of CiviCRM. However the type it defines is not declared yet.

In the solution is to define a type for civicrm_in_operator that is exactly the same as in_operater and add it to the configuration.


mlutfy:
Hi @kainuk, sorry for the slow response, but for Drupal8 we should aim to remove the views integration from CiviCRM core and instead rely on entity_civicrm (see: https://lab.civicrm.org/dev/drupal/issues/39).

PR #27, opened 2019-08-21
dev/drupal#83 Disable qf key validation for forms interact with CMS

sunilpawar:
https://lab.civicrm.org/dev/drupal/issues/83


KarinG:
I just reproduced this - commented on https://lab.civicrm.org/dev/drupal/issues/83

KarinG:
I've reproduced this issue on a live site as well as on a local. Applying this PR fixes the issue described on https://lab.civicrm.org/dev/drupal/issues/83 -  with this PR I can create a user accounts regardless of whether a CiviCRM Profile is configured for: "Drupal User Registration". In addition this makes sense.

eileenmcnaughton:
Thanks for the speedy review @KarinG and for the fix @sunilpawar

PR #32, opened 2019-10-29
Add CiviCRM icon to Toolbar

homotechsual:
## Overview
_This PR alters the current invocation of `hook_toolbar()` to pull in a small CSS file to display the CiviCRM icon on the Drupal 8 toolbar. This is important for mobile compatibility where the Toolbar is shrunk to icons only and the CiviCRM item currently disappears or reverts to text only mode. Also used when the toolbar is in an alternative location (e.g: left aligned)._

## Before
_No icon on toolbar_
Before screengrab.

## After
_CiviCRM icon on toolbar_
After screengrab


mlutfy:
Nice improvement :)

Question about the icons: were they copied from another repo? Would it complicate things if they were in civicrm-core?

css/civicrm.icons.css
```diff
@@ -0,0 +1,37 @@
+/**
+ * @file
+ * Styling for CiviCRM toolbar icons.
+ */
+
+/* Bar Icons */
+.toolbar-bar .toolbar-icon-civicrm:before {
+    background-image: url("../icons/bebebe/CiviCRM.svg");
+  }
```
mlutfy: Could you fix the indentation? There should be two spaces, and the closing curly bracket should be aligned with the parent block.

homotechsual:
> Nice improvement :)
> 
> Question about the icons: were they copied from another repo? Would it complicate things if they were in civicrm-core?

They were custom creations from the new white icon from the `marketing-assets` repo on Lab. The Drupal toolbar uses a specific set of colours which I matched for consistency - they could indeed live in any repo as long as the path is accessible to CSS/browser.

The current icon folder structure is `icons/color/CiviCRM.svg|.png`

css/civicrm.icons.css
homotechsual: Done a coding standards pass on the CSS.

homotechsual:
Mobile screenshots (simulated!):

**Before:**
Before on mobile


**After:**
After on mobile

mlutfy:
I did a bit of live testing, although admittedly on an old version of civicrm/drupal, but is this behaviour expected:

civicrm-icon-Peek 29-10-2019 16-57

i.e. the mobile icon is green, whereas the desktop one is monochrome.

I'm using adminimal with adminimal's menu. If that's not what you're seeing, I'm OK with blaming my environment and merge.

homotechsual:
I'm seeing the same. Need to test this pre-patch as well.

css/civicrm.icons.css
```diff
@@ -0,0 +1,37 @@
...
+.toolbar-bar .toolbar-icon-civicrm:active:before,
+.toolbar-bar .toolbar-icon-civicrm.is-active:before {
+  background-image: url("../icons/ffffff/CiviCRM.svg");
+}
+.no-svg .toolbar-bar .toolbar-icon-civicrm:active:before,
+.no-svg .toolbar-bar .toolbar-icon-civicrm.active:before {
+  background-image: url("../icons/ffffff/CiviCRM.png");
+}
+  
+
```
mlutfy: There seems to be a trailing whitespace on these two lines.

homotechsual:
This is a pre-patch site:

image

homotechsual:
The CSS/icon for this are coming from the CiviCRM menu trying to collapse for mobile on Drupal 8 - it's a bit broken and we need a better solution to this overall - I'll find some time to take a look at it in the next week or two.

This only happens when the CiviCRM menu item is "active" - but the behaviour exists independent of this PR.

css/civicrm.icons.css
homotechsual: Fixed.

mlutfy:
After a bit of discussion online, we agreed that this is OK to merge, and the mobile menu icon is a separate core issue.

homotechsual:
Mobile menu icon issue: https://lab.civicrm.org/dev/core/issues/1349

PR #36, opened 2020-01-30
(WIP) dev/drupal#4 - Exploratory branch using civicrm-setup (5.21)

totten:
When performing installation, use `civicrm-setup` instead of `install/`. This is a more organized/flexible process - and it generates SQL on-demand (instead of requiring pre-generated SQL files).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
   ],
   "require": {
     "composer/installers": "^1.0",
-    "civicrm/civicrm-core" : ">=5.4.0"
+    "civicrm/civicrm-core" : ">=5.4.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -27,7 +27,8 @@
   ],
   "require": {
     "composer/installers": "^1.0",
-    "civicrm/civicrm-core" : ">=5.4.0"
+    "civicrm/civicrm-core" : ">=5.4.0",
+    "civicrm/civicrm-setup" : "~0.4.1"
```
seamuslee001, line 31: @totten I'm guessing this will need to be updated now

totten:
I believe this was previously merged via #37.

PR #37, opened 2020-01-30
dev/drupal#4 - Switch automatic installation to use the civicrm-setup API

totten:
Overview
---------

In the Civi-D8 module, the CiviCRM installation can be done automatically. (*It doesn't have to be automatic -- one can use `cv core:install` for more advanced installation options. But that's not the main focus of this PR.*)

The automatic installation relies on functionality in core to do the installation. This PR switches from older core functions to newer core functions. In the process, it addresses a structural issue involving SQL schema and `composer`-based sites.

See also:

* https://lab.civicrm.org/dev/core/issues/1615
* https://lab.civicrm.org/dev/drupal/issues/4

Before
-------

* `hook_install` performs an automatic installation by calling `$civicrm_base/install/civicrm.php`.
* `hook_requirements` uses `Civi\Install\Requirements.php` to check pre-install requirements.
* You must have various `*.mysql` files in the `civicrm-core` tree. These files are generated via `GenCode` and are not stored in `civicrm-core.git`. This is a key reason why the `roundearth/civicrm-composer-plugin` (for Civi-D8) downloads an extra copy of the Civi-D7 tarball.

After
-----

* `hook_install` performs an automatic installation by calling the [Setup API](https://github.com/civicrm/civicrm-dev-docs/pull/765/)
* `hook_requirements` uses Setup API to check pre-install requirements.
* SQL schema are generated on the fly - you don't need to have `*.mysql` files.

QA / Testing: Installation
-----------------

The main issue is exploring different installation scenarios. Some of these are covered by automated testing, some aren't. I've tested a considerable number of scenarios during development, and it would be painful for a reviewer to try to do all the same cases, so let me outline the cases I tried (and hopefully a reviewer can direct their energy at the cases I didn't think about).

Consider a QA matrix of different file structures (e.g. `drupal8-clean` or `d8prj-re`) versus activation workflows (e.g. `cv core:install` or `drush en civicrm` or activating via web UI). For each combination, one prepares a new build and spot-checks a few pages in the CiviCRM UI. These are my latest results (circa Feb 26) from testing this branch in each case:

| Workflow                 | (1) `drupal8-clean` | (2) `d8rec-clean` | (3) `d8prj-re` |
| ------------------------ | ------------------- | ----------------- | -------------- |
| (A) Using `cv core:install`    | ok                  | ok                | skip           |
| (B) Using `drush en`     | ok                  | ok                | ok    |
| (C) Using web enable     | ok                  | ok                | ok   |

[...]

PR #38, opened 2020-02-25
dev/drupal#107 - Fix user<=>contact sync used by installation workflow

totten:
Overview
--------

This fixes a bug which primarily manifests for the administrator during a new Civi-D8 installation.

See: https://lab.civicrm.org/dev/drupal/issues/107

Before
------

Two ways to look at this:

(A) __Main symptom__ - After enabling CiviCRM on D8, you navigate to the CiviCRM dashboard and get a weird error like:

> Sorry, due to an error, we are unable to fulfill your request at the moment. You may want to contact your administrator or service provider with more details about what action you were performing
>
> DB Constraint Violation - contact_id should possibly be marked as mandatory for DashboardContact,create API. If so, please raise a bug report.

(B) __Formal description__ - If (for whatever reason) a D8 user is not linked to a Civi contact, then any attempt to work with the active user/contact will fail. (This affects the admin user during installation... but it likely affects *all* users/sessions that exist around the time of installation, and it also likely affects some edge-cases when working with DB snapshots.)

After
-----

(A) __Main symptom__ - After enabling CiviCRM on D8, you can start using CiviCRM. No need to logout/login.

(B) __Formal description__ - If (for whatever reason) a D8 user is not linked to a Civi contact, and if the user tries to use a `civicrm` page, then a contact will be created and linked.

Technical Details
-----------------

This same technique has been active in Civi-D7 for years. (It seems to go back at least as far as the svn=>git migration.)

The code for this technique was present in Civi-D8, but it had some comments indicating that its significance wasn't understood at the time of porting, and it was run at the wrong moment (i.e. *after* the page-execution).

Aside: I've been doing my local testing in hacked-up variant of the [drupal8-clean](https://github.com/civicrm/civicrm-buildkit/tree/master/app/config/drupal8-clean) configuration -- specifically, editing `install.sh` to omit the automated Civi installation and also using #37. However, I believe this would fix the symptom in other configurations.

### DIFF
--- a/src/Controller/CivicrmController.php
+++ b/src/Controller/CivicrmController.php
@@ -69,6 +69,11 @@ class CivicrmController extends ControllerBase {
     // Need to disable the page cache.
     \Drupal::service('page_cache_kill_switch')->trigger();
 
+    // Synchronize the Drupal user with the Contacts database (dev/drupal#107)
+    if (!$this->currentUser()->isAnonymous()) {
+      $this->civicrm->synchronizeUser(User::load($this->currentUser()->id()));
+    }
+
     // @Todo: Enable CiviCRM's CRM_Core_TemporaryErrorScope::useException() and
     // possibly catch exceptions. At the moment, civicrm doesn't allow
     // exceptions to bubble up to Drupal. See CRM-15022.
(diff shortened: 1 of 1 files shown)


demeritcowboy:
* General standards
    * [`r-explain`] __PASS__
      * And I've seen the problem before so know it exists.
    * [`r-user`] __PASS__
    * [`r-doc`] __PASS with comment__
      * For roundearth the docs could now exclude the logout step.
    * [`r-run`] __PASS__
      * Solves the problem.
      * I tested with roundearth because I'm not set up (on windows) for buildkit. It's not a preference thing just..windows.
* Developer standards
    * [`r-tech`] __PASS__
    * [`r-code`] __PASS__
    * [`r-maint`] __PASS__
      * I can't immediately think how to do a test.
    * [`r-test`] __PASS__

mlutfy:
Thanks @totten @demeritcowboy !

PR #41, opened 2020-06-10
Implement Metatags hook to fix page titles

seamuslee001:
…titles are properly rendered in the metatags

ping @karing @bgm @jackrabbithanna 

The issue we had was that on a client's site there was the metatag module installed and the configuration was very standard. e.g. title was set to "[current-page:title] | [site:name]" 

We found that for example on Contribution Pages that even tho the title on the page was correctly set the page title The title in the browser tab and in other metatags was set to `CiviContribute | site name` because the route name used that as the holding title https://github.com/civicrm/civicrm-core/blob/master/CRM/Contribute/xml/Menu/Contribute.xml#L31 and not the page title as set via `CRM_Utils_System::setTitle`


JoeMurray:
Question: should this function be put into a different package than a separate module? This is a fairly popular module https://www.drupal.org/project/usage/metatag and there may be similar problems with similar modules, eg EasyMeta.

KarinG:
@seamuslee001 - thanks for addressing this. I'm a bit worried though that we're going to need to allow others to implement `_alter` for (many) more modules?

seamuslee001:
@KarinG possibly, I would also note that in the example module Joe gave of easyMeta that there is any way for 3rd party modules to affect the content of easyMeta but also the stats suggest that most people are using the Metatag module.

If we didn't feel it was appropriate here perhaps we (JMA) could host this on our own github instead as a way forward. This certainly solved the problem for our client but do want to get community thoughts on how to resolve this cos I am sure that others have probbly run into the same issues right?

mlutfy:
~~I also agree it's a nice addition, and thanks for sharing it, but I feel like it should be a third-party contrib on drupal.org, same as webform_civicrm, roundearth theme, etc. I don't think the maintainers of this repo (civicrm-drupal-8) have the capacity to maintain/review future patches for those modules.~~

mlutfy:
I hit reply a bit too quickly: is there no other way to handle the page title? I agree it's a rather important bug :(

KarinG:
@seamuslee001 - I'm with you in that we need to solve this. I'm just wondering re: the best way. For people to have to enable an extra module seems like that should not be necessary...

seamuslee001:
So the issue seems to be that the token is being filled with the content based on the [route definition](https://github.com/civicrm/civicrm-drupal-8/blob/master/src/Routing/Routes.php#L34) rather than based on the call back in the controller here https://github.com/civicrm/civicrm-drupal-8/blob/master/src/Controller/CivicrmController.php#L95 The route is being set based off of the CiviCRM XML files so for a large number of them its ok but its only when you want to dynamically override the title dynamically using CRM_Utils_System::setTitle e.g. ContributionPage where the route has just CiviContribute as the title https://github.com/civicrm/civicrm-core/blob/master/CRM/Contribute/xml/Menu/Contribute.xml#L31

jackrabbithanna:
That code snippet could probably be in the Core module and not depend on metatag. If you only implement that hook, and Metatag provides the hook, it won't fire unless Metatag is installed. It would be good UX to not need to install a contrib module for 10 lines of code that does the one thing. Don't know if there would be anything else ever wanted to be added to it or not that would require requiring metatag in the .info file.
Submodule in core is better than out in contrib IMO, unless it was to grow more functionality and drift into something really out of scope for what Core Team wants to maintain... 

For the module, the drupal standard would be for the module name to be civicrm_metatags not civicrmmetatags

JoeMurray:
Thanks for the advice @jackrabbithanna - it seems good to me. I like the point that the function won't get called unless Metatag is enabled on the site.

mlutfy:
@seamuslee001 Neat fix, thanks for the details!

mlutfy:
Not a blocker, but minor caveat for eventual follow-up issue: this does not seem to work on AngularJS pages (ex: Mosaico new mailing).

colemanw:
Might wanna reference this for AngularJS pages:

https://github.com/civicrm/civicrm-core/blob/13a3d2142f2600376f73c5b43f3a4db9d3a3e49d/ang/crmUi.js#L1048-L1055

seamuslee001:
@colemanw @mlutfy my guess is that given that AngularJS runs in browser it would have to manipulate the title or similar would be my guess but I'm not 100% sure on that

seamuslee001:
@KarinG @mlutfy @jackrabbithanna I have now shifted this code to the core module as suggested

KarinG:
👍  that looks great - thanks @seamuslee001

mlutfy:
Merging a per review from Karin and Mark.

I've been using this in production on a few sites and it definitely helps.

Thanks @seamuslee001 !

PR #42, opened 2020-07-18
dev/drupal#91 - Require email otherwise contact UF sync doesn't work properly

demeritcowboy:
https://lab.civicrm.org/dev/drupal/-/issues/91 and see also https://github.com/civicrm/civicrm-core/pull/17771

This makes email required for CMS user accounts when using CiviCRM. In drupal 8 it is not required, which has its uses, but doesn't work well with initial UF synchronization because it needs email (technically uniqId, which is email).

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -86,19 +86,19 @@ function civicrm_toolbar() {
         '#options' => [
           'attributes' => [
             'title' => t('CiviCRM'),
-            'class' => array(
+            'class' => [
               'toolbar-item',
               'toolbar-icon',
               'toolbar-icon-civicrm',
-            )
+            ],
           ],
         ],
       ],
       '#attached' => [
         'library' => [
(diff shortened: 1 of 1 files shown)


homotechsual:
Tested on current RC + D8 - confirmed working.

jackrabbithanna:
Sure enough email isn't required by Core, wow. I think making email required when CiviCRM is installed is just fine. I have trouble imagining a scenario where a Drupal user just absolutely DOESN"T need a email address.
Implementation looks fine to me. 
+1

mlutfy:
Merging based on the reviews. Thank you!

eileenmcnaughton:
Thanks for prompt feedback @KarinG @jackrabbithanna @MikeyMJCO

demeritcowboy:
Thanks everybody!

PR #47, opened 2020-09-13
dev/drupal#122 - getUsername() is deprecated

demeritcowboy:
Related to https://lab.civicrm.org/dev/drupal/-/issues/122

Where this is used is on the drupal user account screen in edit mode when you click on the Name and Address tab that civi provides.

The easiest way to see this is to install drupal 9 with civi and then visit that Name and Address tab. It's difficult to see deprecation notices in drupal 8 for reasons I won't go into here.

https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Session%21AccountInterface.php/function/AccountInterface%3A%3AgetUsername/8.9.x

https://git.drupalcode.org/project/drupal/-/blob/3cd84c3f1790b48315168f73db0789a004173a69/core/modules/user/src/Entity/User.php#L376


seamuslee001:
This looks correct to me merging

demeritcowboy:
Thanks!

PR #48, opened 2020-09-16
Update docs links.

homotechsual:
Updates the docs page links and replaces forum link with CiviCRM.SE

### DIFF
--- a/src/Civicrm.php
+++ b/src/Civicrm.php
@@ -29,12 +29,12 @@ class Civicrm {
     }
 
     // Get ready for problems.
-    $docLinkInstall = "http://wiki.civicrm.org/confluence/display/CRMDOC/Drupal+Installation+Guide";
-    $docLinkTrouble = "http://wiki.civicrm.org/confluence/display/CRMDOC/Installation+and+Configuration+Trouble-shooting";
-    $forumLink = "http://forum.civicrm.org/index.php/board,6.0.html";
+    $docLinkInstall = "https://docs.civicrm.org/installation/en/latest/drupal8/";
+    $docLinkTrouble = "https://docs.civicrm.org/installation/en/latest/general/troubleshooting/";
+    $seLink = "https://civicrm.stackexchange.com";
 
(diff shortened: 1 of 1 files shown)


homotechsual:
See also civicrm/civicrm-drupal#624

demeritcowboy:
While the sentiment is right doing an r-run there's a couple things:

The advice given on line 54 where this string gets used isn't really the right advice anymore. If civicrm_root is wrong you don't even get this far because the autoloader itself can't even load. The most likely way to reach that line is something like if CRM/Core/Config.php is missing, which is one way I could trigger it.

Also either t() doesn't work at this stage in the boot process or something else is going on because the error looks like this on screen:

`Drupal\civicrm\Exception\CiviCRMConfigException: The path for including CiviCRM code files is not set properly. Most likely there is an error in the <em>civicrm_root</em> setting in your CiviCRM settings file (!1).civicrm_root is currently set to: <em>!1</em>.Please review the <a href='!1'>Drupal Installation Guide</a> and the <a href='!2'>Trouble-shooting page</a> for assistance. If you still need help installing, you can often find solutions to your issue by searching for the error message in the <a href='!3'>installation support section of the community forum</a>.</strong></p> in Drupal\civicrm\Civicrm->initialize() (line 57 of modules\contrib\civicrm\src\Civicrm.php).`

demeritcowboy:
This is interesting - is `!` no longer used in t() in drupal 8? The code doesn't seem to support it? https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Render%21FormattableMarkup.php/function/FormattableMarkup%3A%3AplaceholderFormat/8.2.x

https://www.drupal.org/node/2575819

demeritcowboy:
It seems they don't want you to use html/links in exceptions. They get escaped and you end up seeing html code. This appears to be by design.

Additionally, using the string concatenator `.` with two t()'s converts the objects to a string, so even if they did let you output html in exceptions it would no longer be a TranslatableMarkup object when it got there.

What does seem to work to output links is to replace the throw with
```
\Drupal::messenger()->addError(...);
return;
```
but depending on what is messed up in the particular install throwing an exception might be the more sure thing to do. But it's probably reasonable to assume the drupal part is working if they got this far, but exception with plain text might be better for command line errors.

So 50/50 on what to use?

And also the `Most likely there is an error in the <em>civicrm_root</em> setting in your CiviCRM settings file` should be something like `Most likely CRM/Core/Config.php is missing or your include_path is not set correctly`.

seamuslee001:
@MikeyMJCO are you able to respond to @demeritcowboy 's comments above ^

PR #50, opened 2020-09-25
Use the new-ish frontend_title field of the profile for the menu title

adixon:
As per https://lab.civicrm.org/dev/drupal/-/issues/132

### DIFF
--- a/src/Plugin/Derivative/LocalTasks.php
+++ b/src/Plugin/Derivative/LocalTasks.php
@@ -37,7 +37,7 @@ class LocalTasks extends DeriverBase implements ContainerDeriverInterface {
     foreach ($uf_groups as $key => $uf_group) {
       if ($uf_group['is_active']) {
         $this->derivatives["civicrm.{$key}"] = $base_plugin_definition;
-        $this->derivatives["civicrm.{$key}"]['title'] = $uf_group['title'];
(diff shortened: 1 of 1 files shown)


src/Plugin/Derivative/LocalTasks.php
```diff
@@ -37,7 +37,7 @@ public function getDerivativeDefinitions($base_plugin_definition) {
     foreach ($uf_groups as $key => $uf_group) {
       if ($uf_group['is_active']) {
         $this->derivatives["civicrm.{$key}"] = $base_plugin_definition;
-        $this->derivatives["civicrm.{$key}"]['title'] = $uf_group['title'];
+        $this->derivatives["civicrm.{$key}"]['title'] = $uf_group['frontend_title'];
```
seamuslee001: @adixon this broadly is heading down the right path but I would suggest doing this instead

```suggestion
        $this->derivatives["civicrm.{$key}"]['title'] = !empty($uf_group['frontend_title']) ? $uf_group['frontend_title'] : $uf_group['title'];
```

adixon: Yes, that's a good idea. I switched the logic around to save that exclamation mark and my logic processing brain.

seamuslee001:
looks good to me

adixon:
Thanks!

PR #51, opened 2020-09-25
Include additional civicrm core resources when building a UserProfile

adixon:
As per https://lab.civicrm.org/dev/drupal/-/issues/113

Allows hook_civicrm_buildProfile to work properly.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+custom: https://civicrm.org/civicrm/contribute/transact?reset=1&id=47
--- a/civicrm.install
+++ b/civicrm.install
@@ -8,6 +8,8 @@
 use Drupal\Core\Database\Database;
 
 /**
+ * Implements hook_install().
+ *
  * Contains things that need to be installed when the module is installed.
  */
 function civicrm_install() {
@@ -27,6 +29,17 @@ function civicrm_install() {
 
   $setup->installFiles();
   $setup->installDatabase();
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function civicrm_uninstall() {
+  /** @var \Civi\Setup $setup */
+  $setup = _civicrm_setup();
+
+  $setup->uninstallFiles();
+  $setup->uninstallDatabase();
 }
 
 /**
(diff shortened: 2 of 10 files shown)


demeritcowboy:
Thanks for the patch just I get `CRM_Core_Exception: Error: addCoreResources only supports html-header in CRM_Core_Resources->addCoreResources() (line 377 in CRM/Core/Resources.php)` when I try it.

demeritcowboy:
If I remove the first line it seems to work though.

For the `instance` render line I guess that only does something if you've added custom stuff in that section? It seems to have blank output on a stock install?

adixon:
Thanks for the testing. The first two lines are solving the problem of hook_civicrm_buildProfile which doesn't work out of the box.

But I haven't tested it on the newest civi version, or on a site that doesn't implement hook_civicrm_buildProfile, so I should do that.

stesi561:
Note this will need rebasing now.
I concur with Demerit, further based on your comment above I removed the first two lines and it worked. 
Tested on:  D9.08 civicrm 5.31.1

stesi561:
Not this results in the Civicrm admin menu showing on the /user page. So we've left this off for now. @jitendrapurohit has got this working via an alternatively mechanism.

VangelisP:
> Not this results in the Civicrm admin menu showing on the /user page. So we've left this off for now. @jitendrapurohit has got this working via an alternatively mechanism.

I'm really interested on finding what's the alternative mechanism is. Currently, I am facing the same issue as @adixon . As @demeritcowboy mentioned, removing the first line resolves the issue.

As for the menu that appears on the /user/x/edit/profilename page, interestingly, I have the same effect on a D7 website with CiviCRM 5.21 & 5.35 (direct links, no overlay) so I don't find it a big deal. 

Same thing happens (with the menu) here as well: https://dmaster.demo.civicrm.org/user/2/edit/name_and_address

adixon:
My attempt to rebase failed, replacing this by PR 65

https://github.com/civicrm/civicrm-drupal-8/pull/65

demeritcowboy:
I think this can be closed since the other one is more up to date.

PR #52, opened 2020-11-03
[#141] Added hook_uninstall and fixed D9 exception.

VladimirAus:
* Addressing issue 141: https://lab.civicrm.org/dev/drupal/-/issues/141
* Incorporates #45

### DIFF
--- a/civicrm.install
+++ b/civicrm.install
@@ -8,6 +8,8 @@
 use Drupal\Core\Database\Database;
 
 /**
+ * Implements hook_install().
+ *
  * Contains things that need to be installed when the module is installed.
  */
 function civicrm_install() {
@@ -22,11 +24,23 @@ function civicrm_install() {
 
   $installed = $setup->checkInstalled();
   if ($installed->isSettingInstalled() || $installed->isDatabaseInstalled()) {
-    drupal_set_message(t("CiviCRM appears to have already been installed. Skipping full installation."));
+    \Drupal::messenger()
+      ->addMessage(t("CiviCRM appears to have already been installed. Skipping full installation."));
   }
 
   $setup->installFiles();
   $setup->installDatabase();
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function civicrm_uninstall() {
+  /** @var \Civi\Setup $setup */
+  $setup = _civicrm_setup();
+
+  $setup->uninstallFiles();
+  $setup->uninstallDatabase();
 }
 
 /**


civicrm.install
```diff
@@ -22,13 +24,25 @@ function civicrm_install() {
 
   $installed = $setup->checkInstalled();
   if ($installed->isSettingInstalled() || $installed->isDatabaseInstalled()) {
-    drupal_set_message(t("CiviCRM appears to have already been installed. Skipping full installation."));
+    \Drupal::messenger()
+      ->addMessage(t("CiviCRM appears to have already been installed. Skipping full installation."));
```
demeritcowboy, line 28: I still think this needs to throw an exception not a message, because in the situation you end up here you would never see this message because drupal will continue on and run into a different problem that throws a different exception, so it's harder to track down that it actually failed here.

civicrm.install
```diff
@@ -22,13 +24,25 @@ function civicrm_install() {
...
 }
 
+/**
+ * Implements hook_uninstall().
+ */
+function civicrm_uninstall() {
+  /** @var \Civi\Setup $setup */
+  $setup = _civicrm_setup();
+
+  $setup->uninstallFiles();
```
demeritcowboy, line 42: Does this still work when civicrm.settings.php isn't writable? I can test it out later if you haven't already.

civicrm.install
VladimirAus: Warning is sufficient?

civicrm.install
VladimirAus: Haven't tested it when not writable.

civicrm.install
demeritcowboy: But when you try actually running it in the situation where this comes up, the message is never displayed because a later exception ends up getting displayed instead. (https://github.com/civicrm/civicrm-drupal-8/pull/45#issuecomment-698982257)

demeritcowboy:
Thanks for the PR the uninstall part works well. When it's not writable you get the same exception as with command line but it's ok because then you just fix permissions and you can try again and it works fine.

Tested UI, `cv`, `drush`.

**But I still think the other message part needs to throw an exception.**

civicrm.install
homotechsual: I agree, this needs to throw an exception.

seamuslee001: @MikeyMJCO see also https://github.com/civicrm/civicrm-drupal-8/pull/53/files

homotechsual: Based on that I think we need to merge #52 and #53 together!

seamuslee001:
@VladimirAus can you please rebase this PR to just focus on the hook_uninstall item

homotechsual:
Closing this in favour of #61 thanks @VladimirAus for getting the ball rolling on this!

PR #55, opened 2021-01-04
Create a base FunctionalJavascript test class

mglaman:
Port the file from WebForm CiviCRM to allow Drupal PHPUnit FunctionalJavascript testing.

This takes the base file I created for webform_civicrm. This would allow CiviCRM Entity to start having tests.

### DIFF
--- a/tests/src/FunctionalJavascript/CiviCrmTestBase.php
+++ b/tests/src/FunctionalJavascript/CiviCrmTestBase.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace Drupal\Tests\civicrm\FunctionalJavascript;
+
+use Drupal\Core\Database\Database;
+use Drupal\FunctionalJavascriptTests\WebDriverTestBase;
+
+/**
+ * Base class for FunctionalJavascript tests.
+ */
+abstract class CiviCrmTestBase extends WebDriverTestBase {
+
+  protected $defaultTheme = 'classy';
+
+  protected static $modules = [
+    'block',
+    'civicrm',
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+    $this->drupalPlaceBlock('page_title_block');
+    $this->drupalPlaceBlock('local_tasks_block');
+    $this->drupalPlaceBlock('local_actions_block');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function cleanupEnvironment() {
+    parent::cleanupEnvironment();
+    $civicrm_test_conn = Database::getConnection('default', 'civicrm_test');
+    // Disable foreign key checks so that tables can be dropped.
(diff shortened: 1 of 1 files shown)


homotechsual:
This looks good to me. Civilint passes and I can't see any issues with the implementation.

jackrabbithanna:
+1 for this.

KarinG:
@seamuslee001 - If you have any questions please do let me know. This is what we discussed in Mattermost -> porting over the basic test infrastructure that Matt created for Webform CiviCRM over to CiviCRM Core D8 so that other Drupal modules can start using it.

KarinG:
Thanks @colemanw 👍

demeritcowboy:
What I like about this is that ANYONE could still use this for their own internal front-end testing even if they don't use Drupal 8/9 as their CMS-of-choice for their real sites or know ~~anything~~ much about Drupal. As long as you're willing to fight with composer (:imp:) to get the files downloaded, you just need to know phpunit and then learn a couple getThis() type of functions. 👍 

I might have suggested "needs docs". I'm willing to take a stab at it just wondering if there are any already? There's https://www.drupal.org/docs/automated-testing/phpunit-in-drupal but it's in a very awkward order and obviously not targeted at civi people.

KarinG:
I've documented here how I run them -> https://civicrm.stackexchange.com/questions/38141/how-can-i-run-the-d8-webform-civicrm-unit-tests-on-my-local/38142#38142

Note Step L -> 

L. Install Chromedriver

https://chromedriver.chromium.org/

and start it

```
/usr/local/bin/chromedriver
To check that it's running:
karins-MBP:bin sysadmin$ ps -ax |grep chromedriver
62821 ttys004 0:00.18 chromedriver
```

KarinG:
PS - it's SUPER easy to write tests -> I added this two days ago: it creates a webform, configures it to create an activity, it submits it, checks the values stored in CiviCRM and then reloads the webform with cid1 and aid query parameters to update the activity, and then checks the values stored in CiviCRM. 

All in <100 lines of easy to follow commands.

https://github.com/colemanw/webform_civicrm/blob/8.x-5.x/tests/src/FunctionalJavascript/ActivitySubmissionTest.php

demeritcowboy:
Those SE docs are pretty good! Close to what I was thinking. It's just I'd like to try to promote it as not just D8/webform/etc. You can use it without knowing how to use/install drupal itself to write tests against contribution forms, whatever. Thinking...

KarinG:
Yes - and if you have a Github Account you can get free testing + screenshots!

KarinG:
E.g. at the bottom of a failed run: for example: https://github.com/colemanw/webform_civicrm/actions/runs/467685179

you can download -> https://github.com/colemanw/webform_civicrm/suites/1782232202/artifacts/34073575 - to get a series of screenshots to see exactly what happened. E.g. yep - successfully created the webform and has the fields I wanted (or not): this is one screenshot - you can` $this->htmlOutput();` as many as you like. 

image

PR #57, opened 2021-01-10
dev/core#2258 Define CIVICRM_CRED_KEYS during drush installation

seamuslee001:
Not sure if this works at all but figure worth doing anyway

ping @totten


totten:
Gonna go ahead and merge - the equivalent patch has worked on all other environments, and we need to get the RC branched.

PR #61, opened 2021-02-19
dev/drupal#141 - Implement uninstall

demeritcowboy:
https://lab.civicrm.org/dev/drupal/-/issues/141

This is identical to @VladimirAus's PR https://github.com/civicrm/civicrm-drupal-8/pull/52 but rebased and just focuses on uninstall.

I had tested it with UI, drush, and cv: https://github.com/civicrm/civicrm-drupal-8/pull/52#issuecomment-722016349

### DIFF
--- a/civicrm.install
+++ b/civicrm.install
@@ -8,6 +8,8 @@
 use Drupal\Core\Database\Database;
 
 /**
+ * Implements hook_install().
+ *
  * Contains things that need to be installed when the module is installed.
  */
 function civicrm_install() {
@@ -27,6 +29,17 @@ function civicrm_install() {
 
   $setup->installFiles();
   $setup->installDatabase();
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function civicrm_uninstall() {
+  /** @var \Civi\Setup $setup */
+  $setup = _civicrm_setup();
+
+  $setup->uninstallFiles();
+  $setup->uninstallDatabase();
 }
 
 /**


homotechsual:
(*CiviCRM Review Template WORD-1.2*)



* General standards
    * ([`r-explain`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-explain)) __PASS__
    * ([`r-user`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-user)) __PASS__: Users would reasonably expect CiviCRM to uninstall when uninstalling.
    * ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) __PASS__: Changelog style documentation needed probably in sysadmin guide to note that this behaviour has changed and uninstalling the CiviCRM module uninstalls CiviCRM.
    * ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) __PASS__: No issues identified on Drupal 8 or 9
* Developer standards
    * ([`r-tech`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-tech)) __PASS__: Implemented in line with other CMSes 
    * ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) __PASS__
    * ([`r-maint`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-maint)) __PASS__
    * ([`r-test`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-test)) __QUERY__: @eileenmcnaughton would you say this needs test coverage - should we test uninstall/install or is that too heavy?
    
    Thanks to @eileenmcnaughton for surfacing this in the dev digest.

eileenmcnaughton:
@MikeyMJCO I think it would be a pretty big lift to add test cover on this so as long as it doesn't seem like fragile code I think it's probably in the too hard basket. I don't actually know how we would tackle it due to the CRM-CMS integration aspects

homotechsual:
Good enough for me Eileen :-)

demeritcowboy:
There is test cover of sorts at https://github.com/civicrm/cv/blob/e9cedec825477ae3b1959b29cfbaea38ce5a4fb4/tests/Command/CoreLifecycleTest.php#L125, but I'm not sure if that runs on a schedule anywhere.

eileenmcnaughton:
Thanks for picking up this review @MikeyMJCO & for the patch @demeritcowboy

demeritcowboy:
Thanks everybody.

PR #62, opened 2021-02-20
dev/core#2387 - Method of preventing cache for blocks doesn't work

demeritcowboy:
Drupal 8 is super-cachey. The intent was clearly to avoid it but the method used doesn't work. Example to see this is as described in ticket - recent items block doesn't update by itself without clearing drupal cache.

https://lab.civicrm.org/dev/core/-/issues/2387


colemanw:
Merging based on review given in GL.

demeritcowboy:
Thanks!

PR #63, opened 2021-03-30
Add FUNDING.yml to display a SPONSOR link that directs to c.o

homotechsual:


homotechsual:
See https://lab.civicrm.org/community/sustainability/-/issues/20

demeritcowboy:
I have no idea how to test this without it being live but it seems identical to the drupal7 one which displays a link in the right sidebar on github and after seeing it I feel like I should donate money to civi. So let's see what happens.

demeritcowboy:
I think that worked.

PR #65, opened 2021-05-21
dev/drupal#113 - Include additional civicrm core resources when building a UserProfile

adixon:
Replaces PR 51


demeritcowboy:
Just noting where this [was at](https://lab.civicrm.org/dev/drupal/-/issues/113#note_59682) is that it works, but would page-footer be better since then it matches the addScript default that people might be more likely to use.

demeritcowboy:
As per ticket discussion using form-bottom as written seems better and as noted will need to document you need to specify that region in addScript.

demeritcowboy:
There's no merge-ready label here, but will leave open for a day or so if anybody has any other comments.

demeritcowboy:
Docs PR: https://lab.civicrm.org/documentation/docs/dev/-/merge_requests/937

PR #66, opened 2021-10-11
Add civicrm_group_roles

jitendrapurohit:
Move https://www.drupal.org/project/civicrm_group_roles to core civicrm.


demeritcowboy:
Following on from chat https://chat.civicrm.org/civicrm/pl/nbesgzxeptdqpfzndr88x8wbre and not much happening here I wonder if should move this to a lab ticket for concept-approval? Or put out a call for maintainers on the drupal project page? I don't have a good sense of the demand for the module or who would maintain it in core.

stesi561:
I've created https://lab.civicrm.org/dev/drupal/-/issues/182 feel free to drastically revise this!

stesi561:
Found this. I think given the feedback in https://lab.civicrm.org/dev/drupal/-/issues/182 we can close this for now anyway! @jitendrapurohit ?

PR #67, opened 2021-10-11
Fix unreplaced template vars in civicrm.settings.php when using drush installer

demeritcowboy:
Overview
See also https://github.com/civicrm/civicrm-joomla/pull/63

e.g. it leaves behind the literal string `%%dbSSL%%` in the DSN.


seamuslee001:
This works for me tho i haven't tried ever installing via drush

PR #71, opened 2022-02-02
Remove unit tests based on older framework

demeritcowboy:
I don't think these tests work anymore because they are based on simpletest. In any case the community has an equivalent installation test that runs regularly.

@mglaman these aren't your tests but you'd probably be the best person to confirm that they are old and simpletest doesn't work anymore.

I came to these trying to find the base class while thinking about https://github.com/eileenmcnaughton/civicrm_entity/issues/348 and there are two in this repo named civicrmtestbase so I was initially confused, so removing this would also remove some confusion.


mglaman:
These are old and don't run. There is the base class added in this module... or maybe it's still in CiviCRM Entity or Webform CiviCRM. Either way, those two modules have a lot of testing infrastructure and items that could be moved to helpers here (maybe?) Or at least provide plenty of examples.

mglaman (approved):

demeritcowboy:
Thanks @mglaman. Yes the base class is still here and still used: https://github.com/civicrm/civicrm-drupal-8/blob/a77a3a36720510bc23e9c51835933af8132a2aba/tests/src/FunctionalJavascript/CiviCrmTestBase.php

I know there are at least two sets of helper functions out there. I don't know if there's value in trying to move them here yet.

PR #78, opened 2023-02-09
update CiviCrmTestBase setup() method declaration to match BrowserTestBase

jackrabbithanna:
Working on a D10 version of CiviCRM Entity
https://github.com/eileenmcnaughton/civicrm_entity/pull/408

Getting error:
`PHP Fatal error:  Declaration of Drupal\Tests\civicrm\FunctionalJavascript\CiviCrmTestBase::setUp() must be compatible with Drupal\Tests\BrowserTestBase::setUp(): void in /home/runner/drupal/web/modules/contrib/civicrm/tests/src/FunctionalJavascript/CiviCrmTestBase.php on line 23`

The function definition has changed:
https://api.drupal.org/api/drupal/core%21tests%21Drupal%21Tests%21BrowserTestBase.php/function/BrowserTestBase%3A%3AsetUp/10

This may break tests for older versions of CiviCRM though? 

Is there a specific branch being used for Drupal 10?

### DIFF
--- a/tests/src/FunctionalJavascript/CiviCrmTestBase.php
+++ b/tests/src/FunctionalJavascript/CiviCrmTestBase.php
@@ -20,7 +20,7 @@ abstract class CiviCrmTestBase extends WebDriverTestBase {
   /**
    * {@inheritdoc}
    */
-  protected function setUp() {
+  protected function setUp() : void {
     parent::setUp();
     $this->drupalPlaceBlock('page_title_block');
     $this->drupalPlaceBlock('local_tasks_block');


demeritcowboy:
I think this is correct. It's the same change made everywhere else.

Webform hasn't hit it because for historical reasons it predates this file and has its own copy and hasn't switched. Carrot uses this file but hasn't got around to drupal 10 yet.

Just going to run a quick check.

demeritcowboy:
Looks good. And actually webform already made this change (by, err, somebody...): https://github.com/colemanw/webform_civicrm/commit/8c7f139921e1c67b93e4aec6fb2cc54b36e0a7e0#diff-e64ab69e47322c53904e42f0aa9382dacdeab06e5e86fb8ec905fc16cdb7b328

PR #79, opened 2023-02-14
[REF] Add support for composer installers v2 which seems to be needed…

seamuslee001:
… for d10

@demeritcowboy @jackrabbithanna any thoughts on this?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     }
   ],
   "require": {
-    "composer/installers": "^1.0",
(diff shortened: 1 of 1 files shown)


demeritcowboy:
I tried it both this way and also just removing the composer/installers line and both work and both put the module in the web/modules/contrib/civicrm folder. I suppose you would need it if you wanted to install this into something very similar to drupal, but it doesn't have the line in its composer.json, and you wanted it to go into the drupal module folder. That would be a weird project, but ok.

PR #82, opened 2023-05-04
Make the `args` route parameter a string, avoiding a TypeError with Drupal 9.5.9

jensschuppe:
Drupal 9.5.9 introduced ([issue](https://www.drupal.org/project/drupal/issues/3277784), [commit](https://git.drupalcode.org/project/drupal/-/commit/a894a04bac8f2cff2339bbfeb7efc7d5413ad241#79885a62ed357d5f1a684244c606e403d58de4f9_69_75)) a change related to route parameters, making the `args` default that *civicrm-drupal-8* adds for every CiviCRM route in `\Drupal\civicrm\Routing\Routes::listRoutes()` a route parameter, that eventually ends up as a translation parameter when translating the route's title in `\Drupal\Core\Controller\TitleResolver::getTitle()`.

As the `args` parameter is an array (and not a string, as route parameters have to be), this causes a `TypeError` when `\Drupal\Component\Utility\Html::escape()`ing the parameter value.

This PR makes the route parameter a string and `explode()`s it only when needed in the route controller and theme preprocessing.

This is rather important as without that fix, CiviCRM is incompatible with Drupal 9.5.9+.

Cc: @demeritcowboy, @mlutfy, @colemanw

Issue raised in the [CiviCRM chat](https://chat.civicrm.org/civicrm/pl/8asb7ygnybn3ubodns1afr7xda) before.

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -349,7 +349,8 @@ function civicrm_theme() {
 function civicrm_preprocess_html(array &$variables) {
   // Get current route name and CiviCRM args parameter.
   $name = \Drupal::routeMatch()->getRouteName();
-  $args = explode('/', \Drupal::routeMatch()->getParameter('args'));
+  $args = \Drupal::routeMatch()->getParameter('args');
+  $args = is_array($args) ? $args : explode('/', $args);
 
   // Get module from route name.
   $segments = explode('.', $name);
--- a/civicrm.module
+++ b/civicrm.module
@@ -349,7 +349,7 @@ function civicrm_theme() {
 function civicrm_preprocess_html(array &$variables) {
   // Get current route name and CiviCRM args parameter.
   $name = \Drupal::routeMatch()->getRouteName();
-  $args = \Drupal::routeMatch()->getParameter('args');
+  $args = explode('/', \Drupal::routeMatch()->getParameter('args'));
 
   // Get module from route name.
   $segments = explode('.', $name);
--- a/src/Controller/CivicrmController.php
+++ b/src/Controller/CivicrmController.php
@@ -56,7 +56,9 @@ class CivicrmController extends ControllerBase {
    * Main controller, passes trough to CiviCRM.
    */
   public function main($args, $extra) {
-    $args = explode('/', $args);
+    // $args should usually be a `string`, but (when upgrading, e.g. from 5.61.0) it may be `array`.
+    $args = is_array($args) ? $args : explode('/', $args);
+
     if ($extra) {
       $args = array_merge($args, explode(':', $extra));
     }
--- a/src/Controller/CivicrmController.php
+++ b/src/Controller/CivicrmController.php
@@ -56,6 +56,7 @@ class CivicrmController extends ControllerBase {
    * Main controller, passes trough to CiviCRM.
    */
   public function main($args, $extra) {
+    $args = explode('/', $args);
     if ($extra) {
       $args = array_merge($args, explode(':', $extra));
     }
--- a/src/Routing/Routes.php
+++ b/src/Routing/Routes.php
@@ -33,7 +33,7 @@ class Routes {
         [
           '_title' => isset($item['title']) ? $item['title'] : 'CiviCRM',
           '_controller' => 'Drupal\civicrm\Controller\CivicrmController::main',
-          'args' => explode('/', $path),
+          'args' => $path,
           'extra' => '',
         ],
         [


jackrabbithanna:
I can replicate an issue. 
This error happens consistently
```
The website encountered an unexpected error. Please try again later.
TypeError: htmlspecialchars(): Argument #1 ($string) must be of type string, array given in htmlspecialchars() (line 432 of core/lib/Drupal/Component/Utility/Html.php).
```
I did see this after applying the patch, but before clearing caches
`TypeError: explode(): Argument #2 ($string) must be of type string, array given in explode() (line 59 of modules/contrib/civicrm/src/Controller/CivicrmController.php).
`
This happens when going to pages such as /civicrm/a#/status

Testing with CiviCRM 5.57 at this moment. PHP8.1

Applying the patch and clearing caches resolves the issue.

src/Controller/CivicrmController.php
```diff
@@ -56,6 +56,7 @@ public static function create(ContainerInterface $container) {
    * Main controller, passes trough to CiviCRM.
    */
   public function main($args, $extra) {
+    $args = explode('/', $args);
```
jackrabbithanna: I think its possible args is not always an array either. May have to check. If array, explode(), if not, do nothing.

demeritcowboy:
@jensschuppe Can you also backport this to 5.62? Even though it's not a civi regression exactly I think we should backport.

src/Controller/CivicrmController.php
jackrabbithanna: Well, got an error after applying the patch, but before clearing caches, so checking if array before using explode may not be necessary. People must clear caches after applying the patch

demeritcowboy: Ah that might be why I couldn't reproduce before. I do see it now on the system status page, just not on most pages.

jackrabbithanna:
I back-tested against 5.57 and Drupal 9.5.8, and no errors, if cache is cleared after applying the patch.

src/Controller/CivicrmController.php
totten: This might make it more robust during an upgrade:

```suggestion
    $args = is_array($args) ? $args : explode('/', $args);
```

(I'm not sure if it's true on typical D9 upgrades, but - in many environments, during an upgrade - there will be a period where you have old data but need to navigate to perform admin tasks.)

totten:
For cross reference:

* Related threads on chat.c.o: [town hall](https://chat.civicrm.org/civicrm/pl/tiyaxjnp67yimc5nk9k3jfupke), [prod-mtce](https://chat.civicrm.org/civicrm/pl/y7r5qey5jff6ixpezkn6h8j16y),  [Drupal](https://chat.civicrm.org/civicrm/pl/8asb7ygnybn3ubodns1afr7xda)
* Related threads on d.o: [3277784](https://www.drupal.org/project/drupal/issues/3277784), [3352384](https://www.drupal.org/project/drupal/issues/3352384)

The concept on this PR makes sense as a work-around for Civi. Merging this to `5.62-rc` and `5.61-stable` is probably the fastest way to get some kind of fix out there.

FWIW, I think a patch on drupal.org would serve more people:

* [Another d.o user reported a similar problem for `webform`](https://www.drupal.org/project/drupal/issues/3352384#comment-15035166) (*with no indication that Civi was involved*). If we've seen 2 modules within the first 24hr of release, then there's probably more.
* If the work-around goes into CiviCRM, then the versions will have pair-wise constraints (e.g. "Drupal 9.5.9" <=> "CiviCRM 5.61.1"). Some folks who wanted to apply an `X.Y.Z` increment on Drupal will be required to do a `X.Y` increment on Civi.  By contrasting, if the fix for 9.5.9's regression goes into 9.5.10, then the pairings remain more flexible.

(EDIT: Two patches are not mutually exclusive. We can add a work-around to 5.61/5.62 and also encourage d.o to do 9.5.10.)

totten:
* Changed target from `master` to `5.62`.
* Added some more guards so that it's more forgiving during upgrade scenarios.
* Confirmed that it also fixes `E2E\Core\ErrorTest` on D9.5.9.

If all that's agreeable, then let's merge.

demeritcowboy:
Thanks everybody. I'm going to merge, but: Ideally if your cms is `Drupal8`, there would be an upgrade script that runs `drupal_flush_all_caches()`, but that could lead to all kinds of havoc if it runs during a civi upgrade step, which might be fun to watch from a distance, but less fun to experience first-hand. So maybe just a notice that YOU MUST CLEAR DRUPAL CACHE?

jensschuppe:
Thanks all for reviewing and completing this PR while over here it was nighttime :smile:

civicrm.module
```diff
@@ -350,6 +350,7 @@ function civicrm_preprocess_html(array &$variables) {
   // Get current route name and CiviCRM args parameter.
   $name = \Drupal::routeMatch()->getRouteName();
   $args = \Drupal::routeMatch()->getParameter('args');
+  $args = is_array($args) ? $args : explode('/', $args);
```
jitendrapurohit, line 353: Looks like this line is causing the latest [test failure](https://github.com/colemanw/webform_civicrm/actions/runs/4901466347/jobs/8759015859?pr=875) in webform_civicrm? 

Drupal Version - 10.0.9

Detailed error message:

````
Deprecated function: explode(): Passing null to parameter #2 ($string) of type string is deprecated in civicrm_preprocess_html() (line 353 of modules/contrib/civicrm/civicrm.module).
civicrm_preprocess_html(Array, 'html', Array)
call_user_func_array('civicrm_preprocess_html', Array) (Line: 285)
Drupal\Core\Theme\ThemeManager->render('html', Array) (Line: 433)
Drupal\Core\Render\Renderer->doRender(Array, ) (Line: 204)
Drupal\Core\Render\Renderer->render(Array) (Line: 158)
Drupal\Core\Render\MainContent\HtmlRenderer->Drupal\Core\Render\MainContent\{closure}() (Line: 580)
Drupal\Core\Render\Renderer->executeInRenderContext(Object, Object) (Line: 159)
Drupal\Core\Render\MainContent\HtmlRenderer->renderResponse(Array, Object, Object) (Line: 90)
Drupal\Core\EventSubscriber\MainContentViewSubscriber->onViewRenderArray(Object, 'kernel.view', Object)
call_user_func(Array, Object, 'kernel.view', Object) (Line: 111)
Drupal\Component\EventDispatcher\ContainerAwareEventDispatcher->dispatch(Object, 'kernel.view') (Line: 168)
Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object, 1) (Line: 74)
Symfony\Component\HttpKernel\HttpKernel->handle(Object, 1, 1) (Line: 58)
Drupal\Core\StackMiddleware\Session->handle(Object, 1, 1) (Line: 48)
Drupal\Core\StackMiddleware\KernelPreHandle->handle(Object, 1, 1) (Line: 106)
Drupal\page_cache\StackMiddleware\PageCache->pass(Object, 1, 1) (Line: 85)
Drupal\page_cache\StackMiddleware\PageCache->handle(Object, 1, 1) (Line: 48)
Drupal\Core\StackMiddleware\ReverseProxyMiddleware->handle(Object, 1, 1) (Line: 51)
Drupal\Core\StackMiddleware\NegotiationMiddleware->handle(Object, 1, 1) (Line: 51)
Drupal\Core\StackMiddleware\StackedHttpKernel->handle(Object, 1, 1) (Line: 686)
Drupal\Core\DrupalKernel->handle(Object) (Line: 19)
````

Does it make sense to safeguard this line with the following? 

      $args = is_array($args) ? $args : explode('/', $args ?? '');

demeritcowboy: @jitendrapurohit see https://github.com/civicrm/civicrm-drupal-8/pull/85

PR #85, opened 2023-05-06
Fix warnings on non-civi pages from recently fixed args param

demeritcowboy:
When the page isn't a civi page, $args is null, so explode'ing it gives an error.

@totten @jensschuppe @jackrabbithanna

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -350,7 +350,6 @@ function civicrm_preprocess_html(array &$variables) {
   // Get current route name and CiviCRM args parameter.
   $name = \Drupal::routeMatch()->getRouteName();
   $args = \Drupal::routeMatch()->getParameter('args');
-  $args = is_array($args) ? $args : explode('/', $args);
 
   // Get module from route name.
   $segments = explode('.', $name);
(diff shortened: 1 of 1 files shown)


KarinG:
Jitendra determined we need this for D10, because #82 causes ->

image

KarinG:
I'll re-run our D10 test as soon as this PR is merged to confirm.

totten:
Patch looks reasonable IMHO. On MM, @heneryh reported it fixing the problem. Merging.

@KarinG, after you run that D10 suite, please post back -- then we can release.

KarinG:
Running now -> D10 and CiviCRM dev-master https://github.com/colemanw/webform_civicrm/actions/runs/4901466347

demeritcowboy:
@KarinG that webform will still fail until this is upmerged since this is 5.62 and that run is against master. I'll upmerge here now and retrigger it there.

KarinG:
Ah - this is not dev-master - doh… Thank you Dave.

totten:
It looks it was merged forward (5.62=>master). Is that hyperlink (`4901466347`) the same for the new (post-merge) test-run?

KarinG:
Yes it is. Here is the specific D10 - dev-master ✅

https://github.com/colemanw/webform_civicrm/actions/runs/4901466347/jobs/8766334752

PR #89, opened 2024-02-16
Issue #194: add core resources to blocks

herbdool:
Fixes: https://lab.civicrm.org/dev/drupal/-/issues/194


demeritcowboy:
This seems ok. I've put merge-ready just because I'm not sure if it might not play nice with some themes, but noting it's already added to the user profile page so probably would have already come up for the backend.

PR #90, opened 2024-04-03
Check route object exists in processes that do not have routes

omarlopesino:
When I use this module in a drush context, I receive the following error:

    Call to a member function getDefaults() on null.

It happens as in civicrm.module, the route defaults are trying to be obtained, but the route does not exist as it is not a page but a backend process.

Please review, thanks!


demeritcowboy:
Hi,
I can't reproduce so just wondering if you can give more details:

Which drush command are you running?
Also I assume you have metatag installed but also which versions of drupal, php, drush do you have?

omarlopesino:
Drupal version: 10.2.3
Drush version:  12.4.3.0
Metatag version: 2.0.0

It happens when I load an entity in any drush process. The easiest way to reproduce it is to load an entity through drush php.

Steps:

1. Run drush php:

```bash
 drush php
```

2. Load an entity:

```php
    $node = \Drupal::entityTypeManager()->getStorage('node')->load(1);
```

demeritcowboy:
Ok thanks. I still can't get it to trigger the hook by just loading a node with drush but if I trigger the hook manually after like

```
drush ev '$n = \Drupal::entityTypeManager()->getStorage("node")->load(1); 
$mtags = \Drupal::service("metatag.manager")->tagsFromEntityWithDefaults($n); 
$c = ["entity" => &$n];
\Drupal::service("module_handler")->alter("metatags", $mtags, $c);'
```
then I can see it's null, and in normal web usage it's a `Symfony\Component\Routing\Route`, so this looks good.

PR #94, opened 2024-06-24
Permission fix proto

jackrabbithanna:
PR for 5.75 branch, duplicate of https://github.com/civicrm/civicrm-drupal-8/pull/93/

### DIFF
--- a/src/CivicrmPermissions.php
+++ b/src/CivicrmPermissions.php
@@ -38,12 +38,10 @@ class CivicrmPermissions implements ContainerInjectionInterface {
   public function permissions() {
     $permissions = [];
     foreach (\CRM_Core_Permission::basicPermissions(FALSE, TRUE) as $permission => $attr) {
-      $title = array_shift($attr);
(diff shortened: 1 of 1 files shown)


demeritcowboy:
Should we be mirroring what was done in drupal 7? https://github.com/civicrm/civicrm-drupal/pull/678/files

I think we just missed doing it for drupal 8.

jackrabbithanna:
@demeritcowboy I think you're right. My patch here was "quick and dirty" for those in need to get their page back to functional. 
I'd also say that if a patch is going into the drupal-7 repo, that first something equivalent should go into the drupal 8 module, as appropriate.

jackrabbithanna:
@demeritcowboy I pushed a commit with an equivalent logic to the Drupal 7 module update

demeritcowboy:
Looks good, thanks. Test fail is a [known issue](https://lab.civicrm.org/dev/drupal/-/issues/197).

PR #96, opened 2024-06-25
backport #94

demeritcowboy:


demeritcowboy:
Noting the same test fails are always happening in 5.74, e.g. https://test.civicrm.org/job/CiviCRM-Manual-Test/158/

PR #97, opened 2024-06-25
(dev/drupal#197) Router - Fix mismatched routes and "AfformRoutingTest"

jackrabbithanna:
Trying totten's patch from this issue: https://lab.civicrm.org/dev/drupal/-/issues/197


jackrabbithanna:
The work in #88 fixes this issue, but there are other issues there. If we want tests to pass, maybe we test and merge this in the meantime.

totten:
@jackrabbithanna Thanks. I think you're right. We don't need the perfect to be the enemy of the good. The bigger cleanup can merge when it's ready.

The tests are passing, and I think your submission implies that a second set of eyes did look over the patch. +1 for merging this flavor now.

PR #98, opened 2024-08-14
Generate URLs faster - another attempt

demeritcowboy:
Another attempt at https://github.com/civicrm/civicrm-drupal-8/pull/88

This one seems to work with the edge cases we were having trouble with:
* cache clear
* weirdo urls like civicrm/report/instance/14 where the parameter is part of the path
* oddballs like the civicrm/upgrade queue runner

I have not yet tested the installer or done any benchmarking on speed. Going to do that next.

There's pros and cons to using Civi::cache instead of Civi::statics for storing the menu items, which is the bulk of the problem. It seems to work with either but at the moment I've used cache.
For the known paths, I'm not sure we need to even cache those if we're caching the menu build, but I've left that in.

@MegaphoneJon @jackrabbithanna

### DIFF
--- a/src/PathProcessor/CivicrmPathProcessor.php
+++ b/src/PathProcessor/CivicrmPathProcessor.php
@@ -12,16 +12,29 @@ use Drupal\civicrm\Civicrm;
 class CivicrmPathProcessor implements InboundPathProcessorInterface {
 
   /**
-   * {@inheritdoc}
+   * If it's not a civi path, just return it as-is.
+   * If we've looked up the path already, return that.
+   * If there's a longest matching path in the list of defined urls, then fudge any parts of the url after that so that we don't get page not found. This is for weirdo urls like civicrm/report/instance/14, where for some reason the 14 is not a standard url parameter like ?id=14.
+   *   If we could eliminate all those weird ones, then this function shouldn't be needed.
+   * We also separately cache the list of defined urls, since it's expensive to build.
+   * If there is no matching path in the list, just return as-is.
    */
   public function processInbound($path, Request $request) {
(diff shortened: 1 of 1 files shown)


demeritcowboy:
I tested the installer - ok.
For speed tests, with a sample size of 1, I'm not too impressed with the speed improvement. Will need to do more benchmarking. I checked and it is hitting the cache, and the number of items is the same with or without the cache, so it seems to be working as intended.

I'm not worried about the jenkins fails here. Pretty sure they're known issues.

demeritcowboy:
jenkins retest this please

MegaphoneJon:
@demeritcowboy I'm seeing a significant improvement with this, in both speed and memory.  Does your site have a bunch of afforms?  Because without this patch, all the afforms are scanned on each page load to find any missing routes.  With this patch, `afform_civicrm_alterMenu` is never called, which is where most of the time savings comes in.

demeritcowboy:
Thanks for looking. I didn't have many afforms.

And just to clarify your comment about "never", the hook does get called when you clear cache, so it will still do the scan, just not every page.

Sidenote: I do notice now that for that afform hook, asset caching makes a difference too, since it calls that hook on every page with or without this patch if you don't have asset caching on.

I've put merge-ready.

seamuslee001:
@MegaphoneJon @demeritcowboy I like this, just wanted to check that the new cache gets flushed appropriately when a new afform is created (I presume the answer is yes)

demeritcowboy:
I just tried creating a searchkit and then a formbuilder for it with a "page route" set. Then clicked save and visited the page. It opened.
I think that's what you're asking?

PR #99, opened 2024-08-21
civicrm.drush.inc - Drop "civicrm-install" (based on old installer)

totten:
Overview
-------------

`civicrm.drush.inc` provides an installation option based on the old installer methods. We generally want to get rid of the old installer (https://lab.civicrm.org/dev/core/-/issues/1615). This PR would drop the old implementation.

Comments
--------------

There are several ways to perform a CiviCRM installation:

1. Enable `civicrm.module` and let `civicrm.install` perform a default installation (using modern Setup API), or
2. Use `cv core:install` to perform an advanced/configurable installation (using modern Setup API), or
3. Use `drush civicrm-install` to perform an advanced/configurable installation (using old installer)
     * This third option is the one affected by the current PR. It is the only option that still uses the old installer methods.

Is it absolutely necessary to drop `drush civicrm-install`? IMHO, no - it could be retrofitted to use the newer API (*as we did in Civi-WP and Civi-D7*). However, that is more work, and my guess is that it's not needed -- that this implementation isn't really used. Evidence?

* Civi-D8/9/10/11 use `composer` for deployment.
* This installer relies on `civicrm_data.*.mysql`, `civicrm_acl.*.mysql`, etc. -- but these files don't exist in `composer`-based deployments (*unless you take special measures*).
* This installer has options like `--tarfile` which don't make sense for composer-based deployments.
* Most of the drush maintenance for D8/9/10/11 moved to https://www.drupal.org/project/civicrm_drush -- and that version doesn't include `drush civicrm-install` either.

Of course, if `drush civicrm-install` is actually important somehow, then we can close this PR and dig into retrofitting.


totten:
Note: test failures are pre-existing

colemanw:
Funny enough, [our documentation](https://docs.civicrm.org/installation/en/latest/drupal/) specifically says *not* to use drush, in fact to uninstall drush before installing Civi!

image

Which is yet another piece of evidence in support of this. Merging.

ufundo:
does this mean we can delete the `install` directory? have had my eyes on that for a while now 👀

demeritcowboy:
https://github.com/civicrm/civicrm-core/pull/29307 - the PR is awaiting a small update

jackrabbithanna:
Its not necessary to remove the drush package before installation. IDK if that is an artifact of bygone times.

colemanw:
@jackrabbithanna would you mind doing an update to that install doc?

PR #101, opened 2024-09-17
Use a static var for initialized guard for Drupal\civicrm\Civicrm 

ufundo:
This changes the `initialized` variable on the class from an instance var to a static var to prevent re-initialisation globally.

Pertinent case - don't reinitialise every time a URL is generated by `Drupal\civicrm\PathProcessor\CivicrmPathProcessor`

Fixes an occurrence of [this bug](https://lab.civicrm.org/dev/core/-/issues/4478) for me and should improve performance (e.g. generating search kit button links)

### DIFF
--- a/src/Civicrm.php
+++ b/src/Civicrm.php
@@ -3,7 +3,6 @@
 namespace Drupal\civicrm;
 
 use Drupal\civicrm\Exception\CiviCRMConfigException;
-use Drupal\Core\Config\ConfigException;
 use Drupal\Core\Session\AccountInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


ufundo:
Pretty sure those fails are unrelated. @totten ?

totten:
@ufundo Yup, I agree those are pre-existing/unrelated failures.

I think this is an improvement. Flagging `merge-ready` in case @demeritcowboy @jackrabbithanna have other feedback.

demeritcowboy:
This seems good. Clicked around a bit, cleared cache, installed an extension that adds routes, ran upgrade - ok.
Maybe a cron run or long-running cli job might need an explicit re-initialize in some cases, but I don't have a specific case in mind. And/or maybe it should use Civi::statics? Not blocking.
Running some other unit tests as a sanity check - if you don't hear back from me they went ok.

PR #102, opened 2025-01-09
Missing route cachability metadata in CiviCRM

gitressa:
- Solves https://lab.civicrm.org/dev/drupal/-/issues/201

### DIFF
--- a/src/CivicrmBreadcrumbBuilder.php
+++ b/src/CivicrmBreadcrumbBuilder.php
@@ -4,6 +4,7 @@ namespace Drupal\civicrm;
 
 use Drupal\Core\Breadcrumb\BreadcrumbBuilderInterface;
 use Drupal\Core\Breadcrumb\Breadcrumb;
+use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Link;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
@@ -30,7 +31,8 @@ class CivicrmBreadcrumbBuilder implements BreadcrumbBuilderInterface {
   /**
    * {@inheritdoc}
    */
-  public function applies(RouteMatchInterface $route_match) {
+  // public function applies(RouteMatchInterface $route_match) {
+  public function applies(RouteMatchInterface $route_match, ?CacheableMetadata $cacheable_metadata = NULL) {
     $route_object = $route_match->getRouteObject();
 
     // No route object is defined, so we can't inspect it.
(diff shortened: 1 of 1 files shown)


gitressa:
The current patch does not work ... I get this at _/admin/reports/dblog_:

> The website encountered an unexpected error. Try again later.
> 
> Error: Call to undefined method Symfony\Component\Routing\Route::addCacheContexts() in Drupal\civicrm\CivicrmBreadcrumbBuilder->applies() (line 41 of modules/contrib/civicrm/src/CivicrmBreadcrumbBuilder.php).
> 
>

src/CivicrmBreadcrumbBuilder.php
```diff
@@ -30,9 +31,14 @@ public function __construct(TranslationInterface $stringTranslation, CivicrmPage
...
    * {@inheritdoc}
    */
-  public function applies(RouteMatchInterface $route_match) {
+  // public function applies(RouteMatchInterface $route_match) {
+  public function applies(RouteMatchInterface $route_match, ?CacheableMetadata $cacheable_metadata = NULL) {
     $route_object = $route_match->getRouteObject();
 
+    // @todo Remove in Drupal 12.0.0, will be added from ::applies(). See
+    //   https://www.drupal.org/project/drupal/issues/3459277
+    $route_object->addCacheContexts(['route']);
```
demeritcowboy: Hi @gitressa it's not the route object you want to use here, it's the $cacheable_metadata. But note it might be null, so you'll need to check if it's null first.

gitressa: Thanks for the feedback. I simply copied the bits from another patch, and not too sure how to proceed ...

Maybe you can add your tip as a suggestion, and then I'll approve it? Thanks!

demeritcowboy:
```suggestion
    if ($cacheable_metadata) {
      $cacheable_metadata->addCacheContexts(['route']);
    }
```

demeritcowboy: Oh I guess you're using the web UI and not git command line? I can add it as a suggestion just it would be good to [squash the commits](https://docs.civicrm.org/dev/en/latest/tools/git/#rebasing-interactive) which can't be done in the web UI. Another option is close and make a new PR.

gitressa: Thanks for the suggestion, which works perfectly. I no longer get any errors with the new code. And yes, I use UI on Github ...

gitressa: It looks like the merger can Squash on commit:

- https://github.blog/open-source/git/squash-your-commits/
- https://www.reddit.com/r/github/comments/1bhnwj0/how_to_squash_all_commits_of_a_pr_via_github_web/

demeritcowboy:
I do see that option here although I don't see it in the core repo. I think it was decided a long time ago not to use it because of some problem it causes.

Also there are some style check fails so I just made another PR to fix those and squash and then faked your name on the commit: https://github.com/civicrm/civicrm-drupal-8/pull/103

demeritcowboy:
Since this is your first contribution here, can you make another PR in the core repo to add yourself to this file for release notes credit: https://github.com/civicrm/civicrm-core/blob/master/contributor-key.yml

gitressa:
Thanks for the help, it's great that this got fixed so fast.

PR #105, opened 2025-04-30
Add enctype to multipart/form-data if civicrm profile on user registration form have file field

sunilpawar:
Overview
form element missing enctype=multipart/form-data if drupal not using any file type field on user registration form.

Before
enctype was missing, or not equal to 'multipart/form-data' , uploaded file not present in CiviCRM file system.

After
Uploaded file set on contact field.


jackrabbithanna:
Confirming this update was necessary for Civi file files on profiles to work in D9+ when the user form does not have the user picture field on it. This is good from my perspective.

PR #106, opened 2025-05-23
remove redundant CIVICRM_UF_HEAD

ufundo:
Follow up to https://github.com/civicrm/civicrm-core/pull/32254

It's at least a version later as @totten suggested.


demeritcowboy:
There's a style error but it's not from the PR and predates this. Will fix separately.

PR #107, opened 2025-05-31
fix style issue

demeritcowboy:


demeritcowboy:
Going to self-merge this. If removing a blank line causes problems then we're all in trouble.

colemanw:
😱

PR #108, opened 2025-05-31
[WIP] See if can make more errors

demeritcowboy:


demeritcowboy:
(Test fail is just style, and we should probably remove the drupal 9 tests anyway.)
I wasn't expecting anything, but for example if you manually introduce a php warning into a cli-only test, like the e2e hooktest, then before this PR it passes anyway, and after it fails.
There'd still be tests like the widgettest that make actual network calls, but this could be a start.

PR #109, opened 2025-06-11
Customizable name in user/edit/profile as in user/edit

samuelsov:
Simple change to allow the use of `hook_user_format_name_alter` for the CiviCRM profile edit page the same way it's possible for user/edit page (e.g. if we want to display the name from CiviCRM).

To test:
- create a `hook_user_format_name_alter` to change the username
- in `user/edit` the hook is applied
- create a profile that is used for Drupal View/Edit:
Screenshot 2025-06-11 at 15-15-57 Profiles CiviCRM Sandbox on Drupal 10
- in the `user/xx/edit/profile/yy` page, the `hook_user_format_name_alter` was not applied

### DIFF
--- a/src/Form/UserProfile.php
+++ b/src/Form/UserProfile.php
@@ -89,7 +89,7 @@ class UserProfile extends FormBase {
     $html .= \CRM_Core_Region::instance('form-bottom')->render('', FALSE);
     \CRM_Core_Resources::singleton()->addCoreResources();
 
-    $form['#title'] = $this->user->getAccountName();
+    $form['#title'] = $this->user->getDisplayName();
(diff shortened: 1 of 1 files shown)


demeritcowboy:
Some style errors that predate when the style police found this repo:

```
 29 | ERROR | Missing @var tag in member variable comment
 34 | ERROR | Missing @var tag in member variable comment
 39 | ERROR | Missing @var tag in member variable comment
```

jackrabbithanna:
This PR makes sense. +1

demeritcowboy:
@samuelsov @jackrabbithanna I've put up https://github.com/civicrm/civicrm-drupal-8/pull/110 for the pre-existing style errors, so let's merge this.

PR #110, opened 2025-06-17
satisfy phpcs - var declarations

demeritcowboy:
The style errors predate when phpcs started being applied to this repo.

* One of the class members is only ever a local var so it can go. I don't see anywhere this class is extended inside or outside this repo, and why would it be since it's the tab on the drupal user for a civi profile.
* Added `@var` docblock for the other two.


demeritcowboy:
Thanks @seamuslee001

PR #113, opened 2025-10-06
add cache tag to pages with civicrm content

ufundo:
I'm investigating a bug with Drupal page caching. I think this may be part of the solution.

More soon...

### DIFF
--- a/civicrm.module
+++ b/civicrm.module
@@ -57,6 +57,10 @@ function civicrm_page_attachments(array &$page) {
     '#markup' => Markup::create($headers),
   ];
   $page['#attached']['html_head'][] = [$markup, 'civicrm-headers'];
+
+  // if we load CiviCRM content on a page, we add a general cache tag
+  // so we can invalidate Drupal's cache when CiviCRM is flushed
+  $page['#cache']['tags'][] = 'civicrm';
 }
 
 /**


jackrabbithanna:
I think the idea of adding a cache tag and invalidating those is a good idea. Something in the right place like
```
  \Drupal::service('cache_tags.invalidator')->invalidateTags(['civicrm']);
```

ufundo:
Right - I think it maybe goes in here https://github.com/civicrm/civicrm-core/blob/9ba1a8f74716e01ad354e6d3db29bff486951bba/CRM/Utils/System/Drupal8.php#L562

`CRM_Utils_System::flush` is only called in a couple of places currently, but I think it _should_ be triggered by  `cv flush` https://github.com/civicrm/civicrm-core/pull/33728

jackrabbithanna:
civicrm_page_attachments() code runs on more than just /civicrm/* pages where the controller generates the html.

In theory any page could get the 'civicrm' cache tag then .. any page where in the processing the civicrm service initialize() is run at some point. 

So yeah, it makes sense that some pages could have out of date attachments, just perhaps alot of times it doesn't matter.

I'm looking at this: https://github.com/civicrm/civicrm-core/blame/9ba1a8f74716e01ad354e6d3db29bff486951bba/CRM/Utils/System/Drupal8.php#L563

I'm not sure the code comment there is accurate anymore. torrance was doing the initial integration before composer was used. Now that composer is used to build the codebase, one compatible version of symfony is decided upon and downloaded. If something needed to change there it likely could.

jackrabbithanna:
I've thought before it would be better if civi css/js was added via a Drupal library and not this page attachments hook.
This methodology for integration was written many moons ago. We could try defining a library, and using hook_library_info_alter()
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21theme.api.php/function/hook_library_info_alter/11.x

Then perhaps the library discovery cache could be cleared when necessary
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Asset%21LibraryDiscoveryCollector.php/function/LibraryDiscoveryCollector%3A%3AclearCachedDefinitions/11.x

This would be appropriate only if most requests with Civi resources generally use the same css/js .. and there's not things that are dynamic per request. Just a thought, could be wrong. Would be great to just be able to add a library to a build array when we want it.

jackrabbithanna:
Don't let my dreaming about Drupal libraries stop this PR though. Cache tag seems a good idea and a much lower hanging fruit idea.

ufundo:
@jackrabbithanna thanks for looking at this.

I was also dreaming about how to load CiviCRM libraries in a more structured cacheable new Drupal kind of way. I think `hook_library_info_alter` is worth exploring - but will be a bit tricky because of all the different params CiviCRM uses for cache keys (see `getL10nJsParams`)

> civicrm_page_attachments() code runs on more than just /civicrm/* pages where the controller generates the html. In theory any page could get the 'civicrm' cache tag then .. any page where in the processing the civicrm service initialize() is run at some point. So yeah, it makes sense that some pages could have out of date attachments, just perhaps alot of times it doesn't matter.

That's actually the point for me - the case I've been contending with is https://lab.civicrm.org/dev/core/-/issues/6137 - where a CiviCRM cache flush has invalidated the asset builder urls that have been cached into a Drupal Webform. I think a lot of time it *does* matter!

So yeh I like the idea of a more thorough restructuring, but in the short term though the combo of this new tag + `\Drupal::service('cache_tags.invalidator')->invalidateTags(['civicrm']);` seems an improvement on the current.

ufundo:
>  I think a lot of time it does matter!

I also think it's currently very hard to determine where it does or doesn't matter, and if in doubt we should invalidate. The cache tag is an improvement on the only current safe alternative, which is dumping the whole Drupal cache.

colemanw:
Looks like a harmless & useful addition.

PR #114, opened 2025-10-12
Increase priority of service civicrmtheme.negotiator to make civicrmtheme work with other theme negotiator services (e.g. domain_theme_switch)

gngn23:
The submodule civicrmtheme enables us to select a CiviCRM admim theme and public theme.

This is done with the service CivicrmThemeNegotiator, `civicrmtheme.services.yml` sets `priority: 0`.

If we have other drupal modules 
* implementing a ThemeNegotiator service with a *higher* priority 
* that service is checked before the civicrmtheme service
* and if this other services "applies" (i.e. it's implementation of `ThemeNegotiatorInterface::applies()` returns TRUE)
* the civicrmtheme service is never asked
* and the selection for CiviCRM admim theme and public theme are ignored.

That happens because
> the first negotiator returning a theme name wins

https://drupal.stackexchange.com/questions/281447/themenegotiator-not-working-for-admin-pages

In our example we used the contrib module [domain_theme_switch](https://www.drupal.org/project/domain_theme_switch) (building on top of [domain](https://www.drupal.org/project/domain) to use a different theme per domain).
That caused:
* CiviCRM backend pages shown with the domain frontend theme (instead of the CiviCRM theme selected via civicrmtheme)
* and because this particular frontend theme used it's own bootstrap javascript
* the Dropdown-Buttons were not working

The same problem would occur with any other ThemeNegotiator service with a *higher* priority than `0`.

As a solution we increased the priority of civicrmtheme service to 11.

### DIFF
--- a/modules/civicrmtheme/civicrmtheme.services.yml
+++ b/modules/civicrmtheme/civicrmtheme.services.yml
@@ -3,4 +3,4 @@ services:
     class: Drupal\civicrmtheme\Theme\CivicrmThemeNegotiator
     arguments: ['@current_user', '@config.factory', '@civicrm']
     tags:
-      - { name: theme_negotiator, priority: 0 }
+      - { name: theme_negotiator, priority: 11 }


demeritcowboy:
I can't resist: https://www.youtube.com/watch?v=4xgx4k83zzc

vurt2:
I am a workmate of @gngn23 and we have this small patch running for over 3 months and it is working fine.

gngn23:
This was merged into master in november and I can see it in the master branch:
https://github.com/civicrm/civicrm-drupal-8/blob/master/modules/civicrmtheme/civicrmtheme.services.yml

But when I install a new Drupal 10 site with CiviCRM 6.9.1 I do not get the change and I don't see the change in tag 6.9.1:
https://github.com/civicrm/civicrm-drupal-8/blob/6.9.1/modules/civicrmtheme/civicrmtheme.services.yml

Is there a reason for that?
Or what do I not understand about the building of the "tag releases"?

demeritcowboy:
This chart might help explain: https://lab.civicrm.org/dev/release

Basically "master" is roughly 2 months behind release, with some fuzziness in the first week of the month since when master switches is not exactly the same day every month or is sometimes delayed for "reasons".

So master right now is 6.11, and on Nov 17 would have been 6.10.

Another way to see it is check the xml/version.xml in the core tree as of the merge date, e.g. it looked like this on Nov 17: https://github.com/civicrm/civicrm-core/blob/9824aa114bee94662877f59503d3147918629c47/xml/version.xml

PR #115, opened 2025-10-13
make way to install CiviCRM when civicrm.settings.php file present

jackrabbithanna:
I'm developing a way to install CiviCRM when there is an existing civicrm.settings.php present. This requires a PR in civicrm-core, I'll provide link shortly

### DIFF
--- a/civicrm.install
+++ b/civicrm.install
@@ -22,7 +22,7 @@ function civicrm_install() {
 
   $installed = $setup->checkInstalled();
   if ($installed->isDatabaseInstalled()) {
-     throw new \Exception("CiviCRM appears to have already been installed. Skipping full installation.");
+    throw new \Exception("CiviCRM appears to have already been installed. Skipping full installation.");
   }
   if ($installed->isSettingInstalled()) {
     $setup->installFiles(FALSE);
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Works with https://github.com/civicrm/civicrm-core/pull/33796

jackrabbithanna:
IMO this is better than just throwing an exception and breaking the site if the civicrm.settings.php file is present. The Drupal module is installed still after the exception, but database tables or other things are not created, so its WSOD on all pages.
I'm working on a way of automating site creation, and having a canned civicrm.settings.php in a repo.
I'm not sure why the original code wrote in this hard stop if civicrm.settings.php. One is not prevented in installing the module.
Its hard for me to imagine a scenario where this is a bad thing. Anybody know?

demeritcowboy:
There was a PR in core a few years ago about preventing accidentally wiping a site so might be related. I don't think it was drupal though. Will try and find it.

demeritcowboy:
No I was thinking of a different thing, but that still might have been the original thinking, and maybe copied from another cms?

jackrabbithanna:
I've modified this PR to fit changes in the Core PR. This sets a new property on the Setup model to true if the civicrm.settings.php file is present, which will skip creation of the it during installation.

demeritcowboy:
Core PR looks merged so let's get this in the same version.

PR #116, opened 2025-11-19
Add permission check on contact being accessed through profile

mmyriam:
**BEFORE:**

User x

   * with Profile edit permission
   * is logged in
   * edits their account (/user/x/edit)
   * changes tabs to access a profile that was exposed to "View/Edit Drupal User Account" (e.g. /user/x/edit/profile/2)
   * changes the user id in the url to another than their own (e.g. /user/y/edit/profile/2)

Can

   * see the username of user y (which could be their email)
   * see a Save button
   * modify user y's data (explaining how is out of my league but this was reported in a security audit)

**AFTER:**

   * User x can only access the profile for user y if they have permission to edit user y's contact record
   * Otherwise access is denied

PR created with @samuelsov


demeritcowboy:
Thanks @mmyriam this should maybe be reported as a security issue but now that it's here, can you rebase this to be against the 6.9 branch?

mmyriam:
@demeritcowboy I tried to rebase but I don't think I did it right... Maybe I should just start over?

demeritcowboy:
Yeah rebasing is not fun. Sometimes easiest to start over.

PR #121, opened 2026-01-29
PHP 8.x fixes

eileenmcnaughton:
I thought about type hinting `$profile` - I'm 90% sure it should be ?int but I left it


src/Form/UserProfile.php
```diff
@@ -63,17 +63,17 @@ public function getFormId() {
...
+  public function buildForm(array $form, FormStateInterface $form_state, ?AccountInterface $user = NULL, $profile = NULL) {
     // Make the controller state available to form overrides.
     $form_state->set('controller', $this);
     $this->user = $user;
 
     // Search for the profile form, otherwise generate a 404.
     $uf_groups = \CRM_Core_BAO_UFGroup::getModuleUFGroup('User Account');
-    if (empty($uf_groups[$profile])) {
+    if (empty($uf_groups[$profile ?? ''])) {
       throw new ResourceNotFoundException();
```
demeritcowboy, line 74: If profile is null, it throws an error, so we could just bail early if profile is "empty"? I mean throw the error earlier?

demeritcowboy: I mean the url is either /user/N/edit or /user/N/edit/profile/M, and the first has a different route (handled by drupal core). The second is handled by https://github.com/civicrm/civicrm-drupal-8/blob/58780e60836b365d44664f50e6589d70c2894ba1/civicrm.routing.yml#L22

seamuslee001: @demeritcowboy what if someone hits `/user/n/edit/profile`  without the last bit

src/Form/UserProfile.php
```diff
@@ -63,17 +63,17 @@ public function getFormId() {
   /**
    * {@inheritdoc}
    */
-  public function buildForm(array $form, FormStateInterface $form_state, AccountInterface $user = NULL, $profile = NULL) {
+  public function buildForm(array $form, FormStateInterface $form_state, ?AccountInterface $user = NULL, $profile = NULL) {
```
demeritcowboy, line 66: @seamuslee001 then $profile is null, so I'm suggesting this instead:

```suggestion
  public function buildForm(array $form, FormStateInterface $form_state, ?AccountInterface $user = NULL, $profile = NULL) {
  if (empty($profile)) {
    throw new ResourceNotFoundException();
  }
```

eileenmcnaughton: Hmm did I open a can of worms the one I was actually hitting was `?AccountInterface` but I spotted that profile could be NULL and that would be an 8.5 deprecation notice in the places I touched 

I'm not too aware of when it would be NULL.

Is it best if I merge your suggestion of take this back to the one var & that can be separate?

demeritcowboy: To be honest I don't see how accountinterface should ever be null either since this is an existing user edit form

eileenmcnaughton: Hmm - so I hit this on a new - not necessarily functional build - although I think the functional issue is apache layer stuff 

```
eileen@laptop:~/dev/d11$ docker compose exec web bash -lc "cd /opt/drupal && php -d memory_limit=512M /usr/local/bin/cv up "
Found CiviCRM database version 6.10.1.
Found CiviCRM code version 6.10.1.
Preparing extension upgrade...
[PHP Deprecation] Drupal\civicrm\Form\UserProfile::buildForm(): Implicitly marking parameter $user as nullable is deprecated, the explicit nullable type must be used instead at /opt/drupal/web/modules/contrib/civicrm/src/Form/UserProfile.php:66
.
Have a nice day.
eileen@laptop:~/dev/d11$ docker compose exec web bash -lc "cd /opt/drupal && php -d memory_limit=512M /usr/local/bin/cv status "
+-------------------------+----------------------------------------------------------+
| name                    | value                                                    |
+-------------------------+----------------------------------------------------------+
| summary                 | 6.10.1 php84r1011 drupal11 linux                         |
| civicrm                 | 6.10.1                                                   |
| cv                      | 0.3.69 (phar)                                            |
| cv plugins              | 2x (alias-cmd, basic-alias)                              |
| php                     | 8.4.17 (cli, docker, other)                              |
| mysql                   | 10.11.15-MariaDB-ubu2204                                 |
| drupal                  | 11.3.2                                                   |
| os                      | Linux 6.5.0-1024-oem (x86_64, docker)                    |
| smarty                  | 5                                                        |
| path: [civicrm.root]    | /opt/drupal/vendor/civicrm/civicrm-core                  |
| path: [civicrm.l10n]    | /opt/drupal/vendor/civicrm/civicrm-core/l10n             |
| path: [cms.root]        | /opt/drupal/web                                          |
| path: [civicrm.log]     | /opt/drupal/web/sites/default/files/civicrm/ConfigAndLog |
| path: extensionsDir     | /opt/drupal/web/sites/default/files/civicrm/ext/         |
| url: CIVICRM_UF_BASEURL | http://localhost:8080                                    |
+-------------------------+----------------------------------------------------------+
```

demeritcowboy: ok so compile-time php wants the `?`
run-time php complains if the `=null` is removed (because these are extra variables we've added to the interface definition that don't exist in the parent so then it wouldn't match)

So adding the `?` seems fine.

And in fact it looks like that's all that's needed, not even an empty check, since drupal handles a missing `{profile}` even before hitting this function.

eileenmcnaughton: @demeritcowboy ok - this alternative https://github.com/civicrm/civicrm-drupal-8/pull/122

JoeMurray:
Note that Drupal Security validated a vulnerability on this url that had turned in a scan one of clients ran prior to letting us deploy: https://www.drupal.org/project/drupal/issues/3445177

PR #123, opened 2026-03-24
cleaner loading for classes / settings

ufundo:
- clean up classloading
- remove error message directing to help page URLs which no longer exist
- remove reliance on $civicrm_root to find CRM/Core/Config - the autoloader already knows where the core code is
- use `bootSettings` to load settings -- allows settings to be loaded from env vars as well as the `civicrm.settings.php` file

depends on https://github.com/civicrm/civicrm-core/pull/35098


ufundo:
test suite fails without the core patch

build with both looking ok: https://test.civicrm.org/job/CiviCRM-Manual-Test/187/console

PR #124, opened 2026-03-31
dev/drupal#204 - constants not constant enough for drupal

demeritcowboy:
> We need to replace these constants with other constants because somebody might return a number from the hook that isn't a constant, so this way we can validate it.

That's how I interpret their reasoning to change these CONSTANTS so that everyone's hook_requirements implementation stops working until you update it.

I haven't run this exact version yet, just a variation with the replaced constants. Will do some more r-run. To test this fully requires simulating some error conditions and doing it in drupal 10, 11.1, 11.2, and 11.3, since all of those are a little different in this area. The unit tests here will only test 10, and probably only the no-error situation.

https://lab.civicrm.org/dev/drupal/-/work_items/204

### DIFF
--- a/civicrm.install
+++ b/civicrm.install
@@ -51,17 +51,34 @@ function civicrm_requirements($phase) {
 
   $civicrm_base = _civicrm_find_civicrm();
 
+  // Introduced in 11.2, and then compatibility layer removed in 11.3, so we
+  // have to support both.
+  if (class_exists('\Drupal\Core\Extension\Requirement\RequirementSeverity')) {
+    $severityMap = [
+      'info' => \Drupal\Core\Extension\Requirement\RequirementSeverity::OK,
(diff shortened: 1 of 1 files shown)


seamuslee001:
ping @jackrabbithanna

demeritcowboy:
Have done some testing. I'm happy with it.

And just to clarify this isn't just a deprecation, in 11.3+ it prevents the requirements check from working during a UI install and so blocks the install but with no messages - it just seems like nothing happens.

colemanw:
@demeritcowboy is this "severe" enough to go against the RC?

demeritcowboy:
Yeah makes sense.

demeritcowboy:
jenkins test this please

jackrabbithanna:
Hey everyone, I saw this but didn't have time to review and test until today. Anyways, works for me and the things I'm doing. Thank you!

PR #126, opened 2026-04-30
Initial Drupal 12 support

jackrabbithanna:
I have a working local D12/CiviCRM with: 
https://github.com/civicrm/civicrm-core/pull/35569
https://github.com/civicrm/civicrm-core/pull/35570
https://github.com/civicrm/civicrm-core/pull/35571

This PR
And a composer-compile-lib fork that just allows Symfony 8 - PR coming to that repo

See also https://lab.civicrm.org/dev/drupal/-/work_items/204


jackrabbithanna:
@demeritcowboy 
I found that I could not install any other modules after CiviCRM was installed. 
A couple of change records 
install requirements need to be in its own object in D12
https://www.drupal.org/node/3492429
Runtime requirements need to be in the object oriented hook
https://www.drupal.org/node/3490851

I found that if hook_requirements() or hook_runtime_requirements() are present in civicrm.install, installing other modules fails. 

This will be breaking changes now, so maybe we give thought to a separate branch / version for D12. 
This all should be compatible with 11.2+ and D12

src/Hook/CivicrmRequirementsHook.php
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Implements hook_runtime_requirements().
+   */
+  #[Hook('runtime_requirements')]
+  public function runtime(): array {
+    $requirements = [];
+    // modified version of _civicrm_find_civicrm() from civicrm.install
+    // Move to a service?
+    $possible_paths = [];
```
jackrabbithanna, line 22: Move to a service? This code is also needed in CivicrmInstallRequirements

src/Hook/CivicrmRequirementsHook.php
```diff
@@ -0,0 +1,98 @@
...
+        'title' => 'CiviCRM location',
+        'severity' => $severityMap['error'],
+        'description' => 'CiviCRM must be installed via composer.',
+      ];
+      return $requirements;
+    }
+
+    // Modification of _civicrm_setup()
+    // Move to a service?
+    if (defined('CIVI_SETUP')) {
```
jackrabbithanna, line 60: Move to a service? This is also needed in CivicrmInstallRequirements

jackrabbithanna:
The PR for composer-compile-lib https://github.com/civicrm/composer-compile-lib/pull/10

jackrabbithanna:
I created #127 to provide a more backward compatible option. It needs to be discussed, when to stop supporting Drupal 9/10, or even < 11.3 .. and be able to update this module to the latest architectural changes. If it was possible to do a separate repo for Drupal 11.3+ or 12+ .. then we could improve the module much more significantly.

demeritcowboy:
My two cents is that we need to support 10 for a bit but for 9 we should have dropped it a while ago.
Secondary note is that the civi tests don't even run on 11 yet (this is a phpunit compatibility issue - https://lab.civicrm.org/dev/core/-/work_items/5944 - that's something I'd be interested in working on at the sprint).

demeritcowboy:
Up to you if you want to leave this open and come back to it when we're no longer supporting drupal 10-ish.

PR #127, opened 2026-05-14
Drupal 12 Support - More BC Alternative

jackrabbithanna:
Alternative PR to https://github.com/civicrm/civicrm-drupal-8/pull/126

Instead of using the OO Install / Update object/plugins, use hooks in civicrm.install. 

Requires that civicrm.install be required in civicrm.module. 

Adds an implementation of hook_runtime_requirements() to show requirements as before the Drupal status page. 

This is less good that the other IMO, but #126 will require a separate branch/repo because it will not be backward compatible with Drupal 9/10


jackrabbithanna:
The change record necessitating the require in the civicrm.module of civicrm.install https://www.drupal.org/node/3554900

civicrm.install
```diff
@@ -118,6 +118,72 @@ function civicrm_requirements($phase) {
...
+  $setup = _civicrm_setup();
+
+  $sections = [
+    'system' => 'CiviCRM: System',
+    'database' => 'CiviCRM: Database',
+    'other' => 'CiviCRM: Other',
+  ];
+
+
+  foreach ($setup->checkRequirements()->getMessages() as $msg) {
```
demeritcowboy: Is this block supposed to be here? In the original hook_requirements this was copied from, it says only do this block if `$phase !== 'runtime'`, but this new hook_runtime_requirements is the equivalent where `$phase == 'runtime'`.

jackrabbithanna: Yeah it was intention, without this bit then the list of requirements in the Drupal status page doesn't show. Stuff like: 

CiviCRM: Database: Can successfully lock and unlock tables
CiviCRM: Database: Connected
CiviCRM: Database: Database supports MySQL triggers

It can be made to be gone though.

demeritcowboy: It looks like it came from here: https://github.com/civicrm/civicrm-drupal-8/pull/100.

An alternative would be something more fine-grained so could control externally what exactly runs when, but is it worth the effort? `shrug`

jackrabbithanna: well I swear those things used to be there, maybe not recently. I was just testing fresh stock 6.14 D10 install and they're not there now. Ok, well then we can also decide if we need the hook_runtime_requirements at all if we don't want those to show on the status page.

demeritcowboy: Ah this is a good point :)

jackrabbithanna: Yep https://github.com/civicrm/civicrm-drupal-8/commit/08cad4cfd2f8b86bb005be1dc6d175905104f743
well, I guess somebody didn't want the requirements listed. its fine, I'll update this MR

jackrabbithanna:
Removed hook_runtime_requirements() implementation because we had removed runtime requirements in the past :) Now this PR is real simple. I'm doing installation tests on various 10/11 versions before I undraft it.

jackrabbithanna:
Tested install with 10.3, 11.3 and 12.x-dev .. working good for me. After successful install additional modules can be installed/uninstalled without issue.

demeritcowboy:
Works for me.

PR #128, opened 2026-05-30
Add ability for Drupal modules to listen to civi generated symfony events in Drupal modules

jackrabbithanna:
1. Why it doesn't work today

  CiviCRM and Drupal each run their own service container with its own event dispatcher:
each
  - Drupal modules register subscribers in *.services.yml tagged event_subscriber. A compiler pass attaches them to
  Drupal's event_dispatcher service. They never see CiviCRM's container.
  - CiviCRM dispatches all its Symfony events on its own dispatcher — Civi\Core\CiviEventDispatcher, reachable via
  \Civi::dispatcher() / \Civi::service('dispatcher') (vendor/civicrm/civicrm-core/Civi/Core/Container.php:142,
  created as boot service at Container.php:673). This object is built deep inside CiviCRM's boot and is entirely
  separate from Drupal's dispatcher.

  There is an existing bridge, but only for hook-style events. CiviEventDispatcher::bindPatterns() lazily attaches a
  delegateToUF listener to any hook_civicrm_* event (CiviEventDispatcher.php:346), which routes into the CMS hook
  system (CRM_Utils_Hook::runHooks, Hook.php:241) and invokes procedural MODULENAME_civicrm_<hook>() functions in
  .module files (CRM/Utils/Hook/DrupalBase.php).

  So today a Drupal module can react to CiviCRM hooks only by writing a procedural function — and it has no path at
  all to the non-hook Symfony events (civi.api.resolve, civi.token.eval, civi.actionSchedule.*, civi.dao.*, API4
  events, and any event an extension broadcasts) or to listen via an idiomatic EventSubscriberInterface class.

2. Approaches considered

**Approach A** — register Drupal subscribers onto Civi's dispatcher

Mechanism: Collect tagged Drupal services, call \Civi::dispatcher()->addSubscriber() after Civi boots

Fidelity: Full — receives the real event object, correct priority interleaving with Civi/extension listeners, can stopPropagation()/mutate

Overhead: Only the subscribed events

**Approach B** — forward Civi events onto Drupal's dispatcher 

Mechanism: Wrap/decorate Civi's dispatcher to re-dispatch() every event on Drupal's event_dispatcher 

Fidelity: Lossy — Drupal listeners run at one fixed point; priority interleaving and stop-propagation semantics get muddy

Overhead: Forwards everything

This PR is Approach A
It mirrors precisely how CiviCRM extensions subscribe to events, so Drupal developers get a familiar getSubscribedEvents() pattern and full event fidelity.

  3. The key constraints that shape the design

  - CiviCRM's container is compiled and cached to disk, so we cannot bake references to live Drupal service objects
  into it (rules out registering via hook_civicrm_container). Registration must happen at runtime, after the
  dispatcher object exists.
  - CiviCRM boots lazily (Drupal\civicrm\Civicrm::initialize()), and also boots in CLI/Drush/queue contexts. We want
  one registration point that covers all of them.
  - hook_civicrm_config is the ideal trigger: it fires on every Civi boot (web + Drush + cron + queue), early but
  after dispatcher.boot exists, and it reaches any enabled Drupal module's procedural MODULENAME_civicrm_config().
  Guard with \Drupal::hasContainer() so pure cv (Civi-only, no Drupal) safely skips.

  4. Edge cases & documented limitations

  - Boot-time events (the container build, hook_civicrm_container, early civi.* boot events) fire before
  hook_civicrm_config returns — too early for this bridge. Those remain the domain of procedural
  MODULENAME_civicrm_container(). I'll document this.
  - Idempotency: hook_civicrm_config can fire more than once and Civi can re-boot in long-running CLI/queue workers;
  the registrar guards per-dispatcher-instance.
  - Pure cv (CiviCRM CLI without Drupal): \Drupal::hasContainer() is false → cleanly skipped.
  - Subscriber statefulness: subscribers are fetched from Drupal's container, so DI works normally; just avoid
  assuming a Drupal request context in CLI/cron firings.


jackrabbithanna:
Testing with example module: https://github.com/jackrabbithanna/civi_subscriber_example
A Drupal service tagged with "civicrm.event_subscriber"
```
  civi_subscriber_example.event_subscriber:
    class: Drupal\civi_subscriber_example\EventSubscriber\CiviExampleSubscriber
    arguments: ['@logger.channel.civi_subscriber_example']
    tags:
      - { name: civicrm.event_subscriber }
 ``` 

Other than the different tag, this is the same as would be done for standard Drupal event subscribers: https://www.drupal.org/docs/develop/creating-modules/subscribe-to-and-dispatch-events#s-my-first-drupal-event-subscriber

Verified this works and creates log messages when CiviCRM symfony events fire  in Drupal 11.3

jackrabbithanna:
Approach B could be like @demeritcowboy showed here: https://gist.github.com/demeritcowboy/7f25dc79733b524db7f0c5aaaddb92a9

seamuslee001:
This looks quite cool but need to check tests work

colemanw:
Approach A is super cool @jackrabbithanna 🚀

jackrabbithanna:
I tested a fresh Civi install with this PR and the functionality in 10.6. Then tested an existing Drupal 11.3 updating the civi module and the functionality, and both working with the example subscriber module. Looking good!

colemanw:
Well done @jackrabbithanna!

